PR #3, opened 2022-03-29
Switch to new Azure SDK and refactor Azure implementation

phillebaba:
This is a larger refactor of the Azure implementation which switches to the new Azure Go SDK. A lot of function names are the same its for the most part that interface that has changed to become more idiomatic Go. The main benefit of switching is that the new SDK does not require ADAL and instead fully supports [azidentity](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity) which will remove some less than ideal code to get MSI working. I took the liberty to do some refactoring while I am at it in the hope that it will be easier for future developers to understand.

Fixes thanos-io/thanos/issues/5172


metalmatze:
Please make sure to signoff your commits with `--signoff`. To fix the current one use: `git rebase HEAD~1 --signoff`. Thanks!

phillebaba:
E2E tests now pass.

phillebaba:
Biggest issue right now is that MSI Resource and User Assigned ID is an either or choice.
https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#NewManagedIdentityCredential

Dont think this will affect functionality but it will affect the config validation.

phillebaba:
So I have verified that MSI resource no longer has to be set when configuring access to the storage account with Azure. It is automatically calculated and can not be configured.

The current blocker right now seems to be the requirement on Go 1.18 but there seems to be progress on that end also getting Thanos upgraded.

bwplotka:
* Do we have ticket for Go 1.18 upgrade somewhere we can link to track this?
* Also we sill needs tests here, which is not trivial cc @kakkoyun who mentioned has some time for it

kakkoyun:
@phillebaba @bwplotka go1.18 is not strictly required for thanos-io/objstore. Where does this requirement come from?

phillebaba:
The requirement comes from the new Azure Go SDK. It requires Go 1.18, if i remember correctly it is because some authentication package uses generics.

kakkoyun:
@phillebaba Ok, got it. For whatever reason, I thought the other way around. Thanks for clarifying. I think after we release an initial version of this package, we can bump the required version and have these changes.

I'm working on the CI tests now for the repo.

phillebaba:
@kakkoyun so I have now rebased this branch and fixed some of the unit tests. I have also deployed a build of this fork in a test environment and verified that things seems to still work. I just cant figure out why CodeQL is failing.

Looking through the changes it seems that there is no real breaking changes other than the `msi_resource` configuration. It cannot be set any more as the SDK will extrapolate this value. Removing the parameter will cause panics in existing Thanos deployments which set this value as it is no longer a configurable parameter.

On top of this I had to upgrade the Go version to 1.18 which can cause issues for both Thanos and other depending projects.

I think it would be good to start reviewing this PR.

bwplotka:
Thanks, I was just looking at this, will check.

The CodeQL might have problem with different Go version? 

image


Retried just in case.

phillebaba:
@bwplotka I fixed the issue. I needed to set the specific Go version for CodeQL.
https://github.com/github/codeql-action/issues/1059#issuecomment-1185036511

Makefile
```diff
@@ -39,7 +39,7 @@ check-docs: $(MDOX)
 
 .PHONY: deps
 deps: ## Ensures fresh go.mod and go.sum.
-	@go mod tidy -compat=1.17
+	@go mod tidy -compat=1.18
```
bwplotka, line 42: Rebase needed 🤗 

https://github.com/thanos-io/objstore/pull/16

providers/azure/helpers.go
```diff
@@ -4,172 +4,64 @@
...
-	if resource == "" {
-		resource = fmt.Sprintf("https://%s.%s", conf.StorageAccountName, conf.Endpoint)
-	}
-
-	msiConfig := auth.MSIConfig{
-		Resource: resource,
+		return containerClient, nil
 	}
 
+	// Use msi for authentication
```
bwplotka, line 53:
```suggestion
	// Use MSI for authentication.
```

bwplotka (approved):
💪🏽 Thanks, LGTM

I wish we had e2e tested this, but we don't have an Azure account for Thanos use I believe.

Ticket for this: https://github.com/thanos-io/objstore/issues/10

Happy to merge this after rebase (:

providers/azure/azure.go
```diff
@@ -42,16 +42,17 @@ var DefaultConfig = Config{
...
-	PipelineConfig     PipelineConfig     `yaml:"pipeline_config"`
-	ReaderConfig       ReaderConfig       `yaml:"reader_config"`
-	HTTPConfig         exthttp.HTTPConfig `yaml:"http_config"`
+	StorageAccountName string `yaml:"storage_account"`
+	StorageAccountKey  string `yaml:"storage_account_key"`
+	ContainerName      string `yaml:"container"`
+	Endpoint           string `yaml:"endpoint"`
+	MaxRetries         int    `yaml:"max_retries"`
+	// Deprecated: Is automatically set by the Azure SDK.
+	MSIResource    string             `yaml:"msi_resource"`
```
bwplotka, line 51: I think we planned to log warning when someone provides it? Let's do that in next PR if needed (:

bwplotka (approved):
Amazing, thanks 💪🏽 

And thanks for caring about compatibility, good job!

PR #5, opened 2022-04-29
Support s3 compactible implementation with vitrual_host_style

masonmei:
The BucketLookup param in minio.Options is not exposed in s3 config. I suppose we should make this params can be configurable for using.

image


bwplotka:
I think we no longer need this - @kakkoyun synchronized the code (:

bwplotka:
Let me know if we are wrong here

PR #11, opened 2022-07-13
Refactor: Clean up of package placement and unneeded code.

bwplotka:
No functional logic changed, but we need to add e2e test still (next PR).

This will create some pain for importers (Thanos, Cortex, Parca etc) as packages moved but let's do it as early as possible to consider 1.0 for this module (: 

### Changes

* Moved all provider code to providers/ dir.
* Removed testutil and runutil packages, since we already import those from efficient-go/tools/pkg/core
* Removed HTTPConfig aliases for Cortex use - that was done to workaround cycling modules, not needed anymore
* Moved clientutil to exthttp and removed dependency cycle.
* Removed e2e dir, which is unused.

Signed-off-by: bwplotka <bwplotka@gmail.com>


providers/swift/swift.go
```diff
@@ -14,15 +14,14 @@ import (
...
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 	"github.com/ncw/swift"
 	"github.com/pkg/errors"
 	"github.com/prometheus/common/model"
-	"gopkg.in/yaml.v2"
-
 	"github.com/thanos-io/objstore"
-	"github.com/thanos-io/objstore/runutil"
+	"gopkg.in/yaml.v2"
```
kakkoyun, line 24: nit: I don't this these imports are properly formatted using `goimports`

kakkoyun (approved):
LGTM.
Only one nit.

providers/swift/swift.go
kakkoyun: I think external dependencies need to be grouped together.

PR #18, opened 2022-08-07
client,providers: add storj DCS as provider option

stefanbenten:

## Changes

- Adding Storj Provider
- Fixed import ordering for some touched files to make them consistent
- Fixed some smaller items in the README.
- Omitted the missing error checks in the cfggen script.

## Verification

- Compiled and ran it locally, worked fine.
- Still working on adding the "Local Test Network" to ensure full functionality.


brancz:
I have practically no experience with storj, but one of the first things it says in the docs is that it’s S3 compatible, so I wonder what a direct integration does better than using the existing S3 integration?

stefanbenten:
Hello @brancz, thank you for looking at this!

The native integration offers several advantages over the S3 compatible gateway.
One of them being performance (cuts out the proxy/gateway) and the other important one being client side encryption rather than have it be done server side.

brancz:
This generally looks good to me. I wonder though if you'd be up for maintaining this, as if there are problems I don't think any of us have enough knowledge about Storj to be helpful in a meaningful way.

stefanbenten:
Thank you a lot @brancz !

I am more than happy to maintain the integration, let me know if there is anything further needed!

providers/storj/storj.go
```diff
@@ -0,0 +1,207 @@
...
+
+// Close closes the Bucket.
+func (b *Bucket) Close() error {
+	return b.project.Close()
+}
+
+// Iter calls f for each entry in the given directory (not recursive). The argument to f is the full
+// object name including the prefix of the inspected directory.
+func (b *Bucket) Iter(ctx context.Context, dir string, f func(string) error, options ...objstore.IterOption) error {
+	_ = level.Debug(b.logger).Log("Listing %s from Storj Bucket", dir)
```
GiedriusS, line 91: Maybe we can remove these `level.Debug(...).Log()` calls? They would be quite spammy.

scripts/cfggen/main.go
```diff
@@ -70,19 +72,19 @@ func main() {
 
 	errLogger := level.Error(log.NewLogfmtLogger(log.NewSyncWriter(os.Stderr)))
 	if _, err := app.Parse(os.Args[1:]); err != nil {
-		errLogger.Log("err", err)
+		_ = errLogger.Log("err", err)
```
GiedriusS, line 75: It's not necessary to check for `err` [here](https://github.com/thanos-io/objstore/blob/main/.errcheck_excludes.txt#L1) in Log() calls.

GiedriusS (commented):
Are there any containers that we could run to start a new storj object storage? Or is it a proprietary product? It would be ideal to have some integration tests.

scripts/cfggen/main.go
stefanbenten: This was mainly a thing that my environment complained about. I saw it wasnt needed after all, but thought it might be a little cleaner anyway.

PR #19, opened 2022-08-11
Log deprecation notice when MSI Resource is configured for Azure

phillebaba:

## Changes

This change adds a log when the MSI Resource is set in the Azure configuration. The purpose is to make end users aware of the deprecation so that they stop configuring this parameter.

## Verification

Built a fork that I tested in my own cluster.


fpetkovski (approved):
Thanks 👍

bwplotka (approved):

bwplotka:
Thank You!

PR #25, opened 2022-09-01
S3: Support specifying S3 storage class

juanrh:

## Changes

For https://github.com/thanos-io/thanos/issues/5663, enable specifying S3 storage class by specifying it in config.put_user_metadata. For example launching the sidecar with a `--objstore.config-file` as follows stores all S3 files with storage class of ` STANDARD_IA`

```yaml
type: S3
prefix: thanos-test-standard-ia
config:
  endpoint: s3.us-east-1.amazonaws.com
  region: us-east-1
  bucket: MY_BUCKET
  put_user_metadata:
    X-Amz-Storage-Class: STANDARD_IA
  trace:
    enable: true
```

while using the following config file uses the default storage class of `STANDARD`

```yaml
type: S3
prefix: thanos-test-standard-ia
config:
  endpoint: s3.us-east-1.amazonaws.com
  region: us-east-1
  bucket: MY_BUCKET
  trace:
    enable: true
```

## Verification

Added unit tests.

Manual tests launching the sidecar with the two configurations listed above, doing `rm ${PROM_ROOT}/data/thanos.shipper.json` before that to force uploading files, and with a prometheus instance with small block duration to force upload.

```go
${PROM_ROOT}/prometheus --config.file=${PROM_ROOT}/prometheus.yml \
  --storage.tsdb.min-block-duration=1m \
  --storage.tsdb.max-block-duration=1m
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 
 ### Added
 - [#15](https://github.com/thanos-io/objstore/pull/15) Add Oracle Cloud Infrastructure Object Storage Bucket support.
+- [#25](https://github.com/thanos-io/objstore/pull/25) S3: Support specifying S3 storage class.
 
 ### Changed
 
--- a/providers/s3/s3.go
+++ b/providers/s3/s3.go
@@ -96,6 +96,9 @@ const (
 	// NOTE: we're using a context value only because it's a very specific S3 option. If SSE will
 	// be available to wider set of backends we should probably add a variadic option to Get() and Upload().
 	sseConfigKey = ctxKey(0)
+
+	// Storage class header.
+	amzStorageClass = "X-Amz-Storage-Class"
 )
 
 var DefaultConfig = Config{
@@ -160,6 +163,7 @@ type Bucket struct {
 	client          *minio.Client
 	defaultSSE      encrypt.ServerSide
 	putUserMetadata map[string]string
+	storageClass    string
 	partSize        uint64
 	listObjectsV1   bool
 }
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 
 ### Added
 - [#15](https://github.com/thanos-io/objstore/pull/15) Add Oracle Cloud Infrastructure Object Storage Bucket support.
+- [#25](https://github.com/thanos-io/objstore/pull/25) Thanos Objstore S3: Support specifying S3 storage class.
```
bwplotka, line 17:
```suggestion
- [#25](https://github.com/thanos-io/objstore/pull/25) S3: Support specifying S3 storage class.
```

providers/s3/s3.go
```diff
@@ -310,12 +314,23 @@ func NewBucketWithConfig(logger log.Logger, config Config, component string) (*B
 		return nil, errors.Errorf("Initialize s3 client list objects version: Unsupported version %q was provided. Supported values are v1, v2", config.ListObjectsVersion)
 	}
 
+	var storageClass string
+	for _, k := range []string { amzStorageClass, strings.ToLower(amzStorageClass) } {
```
bwplotka, line 318: I am not sure what we are trying to accomplish here 🤔 Do we need this loop?

bwplotka: Ok, I had to read through the whole issue to understand this. Can we:

1. Comment why we have to do it.
2. Can we ensure we are prone to capitalizations? e.g. if user would put `X-Amz-Storage-class` we would miss it. Not a big deal but might be confusing for ppl.

For example instead we could iterate over `PutUserMetadata` and `string.ToLower` for both `amsStorageClass` and map key to check if we are matching it.

3. I wonder if we should not just add `StorageClass` option and fail if user pass it through `PutUserMetadata`. I guess less option is better,so maybe it's fine like this.

bwplotka (approved):
Thanks! Proposed some changes, otherwise LGTM (:

providers/s3/s3.go
juanrh: I implemented item 2, and updated `TestParseConfig_CustomStorageClass` to check different cases of `X-Amz-Storage-Class` are all detected as the same.  

I didn't add a comment for item 1, because I though the new code is more obvious, but I can add another one of you think its needed.   

For item 3, I understood we agreed that not adding a new option was ok, but I'm open to adding a new one if you think that's best. An advantage of reusing `put_user_metadata` is that it doesn't require coming up with a new configuration setting that would make sense for other object stores

matej-g: Thanks for the clarification @juanrh, I think it is fine then, more of a nit from my side. It was not clear to me e.g. why we need to delete that metadata entry but understood it once I went through the issue.

matej-g (approved):

PR #32, opened 2022-10-25
Swift: Allow authentication via application credentials

aarnq:

## Changes

Adds the option to authenticate with application credentials when using Swift.

## Verification

Built and tested with a local version of Thanos verifying using the `tools bucket` commands, and set up a full dev deployment that is currently running.


FUSAKLA:
Hi, great :+1: from me, thanks!

Just one more thing, the configuration options are in [the docs](https://thanos.io/tip/thanos/storage.md/#openstack-swift). I believe this is generated dynamically, so you should be able to update it by running `make docs` IIRC :thinking:

aarnq:
> Hi, great +1 from me, thanks!
> 
> Just one more thing, the configuration options are in [the docs](https://thanos.io/tip/thanos/storage.md/#openstack-swift). I believe this is generated dynamically, so you should be able to update it by running `make docs` IIRC thinking

Thank you for taking a look at it!

I have done those steps, it at least updated the readme in this repo. I assume that is later used in the main repo for `storage.md`. Do you want me add more to it about how to authenticate with application credentials?

FUSAKLA:
In the next PR to bump the dependency on this package, you'll need to run the docs update also, I think.
Regarding documentation of the fields, other fields are not documented either, and I think it's quite self-explanatory and easy to find in Swift documentation.

From me, it is :+1: thanks :tada: 

I do not have the rights to merge this, so maybe @bwplotka could help? :pray:

squat (approved):

FUSAKLA:
Thanks @squat !

PR #34, opened 2022-11-08
azure: Fix ignored options when creating shared credential Azure client

phillebaba:

## Changes

Set client options when creating a Azure client with shared key. Before this the option was only set to clients using MSI authentication.

## Verification

Should not require verification as the same option is passed to the MSI client.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 
 ### Fixed
 - [#33](https://github.com/thanos-io/objstore/pull/33) Tracing: Add `ContextWithTracer()` to inject the tracer into the context.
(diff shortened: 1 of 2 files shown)


phillebaba:
Fixes thanos-io/thanos/issues/5874

phillebaba:
@ahurtaud  could you build a fork with this branch of objstore and verify that it works for you?

ahurtaud:
Sure I should be able to do this tomorrow. Otherwise on monday ;)

GiedriusS (approved):
:+1:

ahurtaud:
@phillebaba I build thanos main + your objstore fix/azure-shared-key-opt branch and it works (tested thanos store only but I guess it'll work for all)
ready to merge.
thank you for the very quick fix!

PR #35, opened 2022-11-09
azure: Update Azure SDK and fix breaking changes

phillebaba:

## Changes

The Azure Go SDK team has once again introduced breaking changes to the SDK. As far as I can see these are mostly related to user experience and have little to do with performance. This change updates all Azure dependencies and fixes the breaking changes to avoid having to make even larger changes in the future.

## Verification

Running tests in my own environment.


providers/azure/azure.go
```diff
@@ -241,54 +234,39 @@ func (b *Bucket) IsObjNotFoundErr(err error) bool {
...
+		Range: httpRange,
 	}
-	resp, err := blobClient.Download(ctx, downloadOpt)
+	resp, err := blobClient.DownloadStream(ctx, downloadOpt)
 	if err != nil {
 		return nil, errors.Wrapf(err, "cannot download blob, address: %s", blobClient.URL())
 	}
-
-	retryOpt := &azblob.RetryReaderOptions{
-		MaxRetryRequests: b.maxRetryRequests,
```
phillebaba, line 270: Need to find where this should now be set.

providers/azure/azure.go
```diff
@@ -318,15 +293,8 @@ func (b *Bucket) Exists(ctx context.Context, name string) (bool, error) {
 // Upload the contents of the reader as an object into the bucket.
 func (b *Bucket) Upload(ctx context.Context, name string, r io.Reader) error {
 	level.Debug(b.logger).Log("msg", "uploading blob", "blob", name)
-	blobClient, err := b.containerClient.NewBlockBlobClient(name)
-	if err != nil {
-		return err
-	}
-	opt := azblob.UploadStreamOptions{
-		BufferSize: 3 * 1024 * 1024,
-		MaxBuffers: 4,
```
phillebaba, line 327: Need to figure out what these are called now. Additionally the values should be documented to specify why they are being set in the first place.

phillebaba (commented):
Most of the changes are just moving code between packages, but there are two issue which need to be resolved before this can be merged.

providers/azure/azure.go
phillebaba: The values have been renamed BlockSize and Concurrency. I think that in a future PR we should consider making these configurable or explaining why this is a good value.

providers/azure/azure.go
phillebaba: So it looks like the new SDK creates issues with reading streams with retries. Keeping this functionality would force reading all the data into memory, something that we have already tracked as a bug.

Looking at the AWS code it does not seem like it implements such logic for reading? If that is the case I suggest we move along with removing the retry logic here.

phillebaba:
Needed to bump faillint version as panics were occurring with the old version for some reason.

saswatamcode (approved):
Thanks! LGTM as far as I understand. 🙂

squat:
Loos like we need to fix a conflict and then we can merge

squat:
(trying to re-trigger tests)

phillebaba:
@squat could we merge this now?

bwplotka:
Thanks!

PR #41, opened 2023-01-13
s3: add support for session token

### LINKED ISSUE #17: S3 Provider doesn't expose a way to pass in the SessionToken
:wave:  I'm trying to use the objstore with S3, and need to provide the sdk with a session token. 

I'm attempting to use the `NewBucketWithConfig` and using dynamic credentials but the config object only exposes the 
`AccessKey` and `SecretKey` but not the `SessionToken` is this something that could be added?

I'm happy to submit the PR but just wanted to see if there was a reason that was left out of the config?

farodin91:
Fixes #17 

## Changes

* s3: add support for session token


matej-g (approved):
LGTM, thanks!

farodin91:
@matej-g does this pr gets merged any time soon? Thanks

matej-g:
@farodin91 sorry for the delay, this looks good to me and since there were no objections / more reviews, I'm merging. Thanks for your contribution!

PR #43, opened 2023-01-27
filesystem: abort filesystem bucket operations if the context has been cancelled

charleskorn:

## Changes

Changed all of the `filesystem.Bucket` methods to return early if the context has been cancelled.

This makes testing cancellation behaviour of consuming applications easier (see https://github.com/grafana/mimir/pull/4061 for an example), and brings the behaviour of the filesystem bucket into line with the other providers.

## Verification

I've added tests for cancellation behaviour in `filesystem_test.go`, and the existing acceptance tests should catch any regressions in existing behaviour.

### DIFF
--- a/providers/filesystem/filesystem.go
+++ b/providers/filesystem/filesystem.go
@@ -53,10 +53,8 @@ func NewBucket(rootDir string) (*Bucket, error) {
 // Iter calls f for each entry in the given directory. The argument to f is the full
 // object name including the prefix of the inspected directory.
 func (b *Bucket) Iter(ctx context.Context, dir string, f func(string) error, options ...objstore.IterOption) error {
-	select {
-	case <-ctx.Done():
+	if ctx.Err() != nil {
 		return ctx.Err()
-	default:
 	}
 
 	params := objstore.ApplyIterOptions(options...)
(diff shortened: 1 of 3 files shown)


providers/filesystem/filesystem.go
```diff
@@ -53,6 +53,12 @@ func NewBucket(rootDir string) (*Bucket, error) {
 // Iter calls f for each entry in the given directory. The argument to f is the full
 // object name including the prefix of the inspected directory.
 func (b *Bucket) Iter(ctx context.Context, dir string, f func(string) error, options ...objstore.IterOption) error {
+	select {
```
pracucci, line 56: [nit] I think you can simplify this into:

```
if ctx.Err() != nil {
  return ctx.Err()
}
```

charleskorn: Good point, resolved in 666b242.

pracucci (approved):
FWIW, LGTM :)

douglascamata (approved):
LGTM ✅

yeya24 (commented):
Thanks!

yeya24 (approved):

PhilipGough (approved):

PR #44, opened 2023-02-02
Add new metric to count total number of fetched bytes from bucket

thib-ack:

## Changes

We are using a cloud provider which bills us on the outgoing trafic, so it is useful for me to know how many bytes are downloaded from the bucket.
I thought maybe others can have the same needs.

Of course if you think that could be useful but want a different implementation, let me know.

## Verification

updated `TestDownloadUploadDirConcurrency` test in `objstore_test.go` but I can add more tests if you want.

Thanks,


matej-g:
Thank you for your contribution @thib-ack 🙇 , I think this change makes sense. Would you mind signing the commits so that we can ensure all the PR checks are green?

thib-ack:
I have signed the commits 🖋️

matej-g (approved):
Thank you @thib-ack! Sorry, looks like this got buried, merging now.

PR #46, opened 2023-02-18
objstore: add experimental encryption wrapper

MichaHoffmann:


MichaHoffmann:
`README.md` changes were automatically created by `make lint`

README.md
```diff
@@ -181,6 +182,9 @@ config:
     encryption_key: ""
   sts_endpoint: ""
 prefix: ""
+encryption:
```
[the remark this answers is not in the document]
bwplotka: `cse` or `client_side_encryption` would work better indeed 👍🏽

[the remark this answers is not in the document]
MichaHoffmann: I went with `client_side_encryption` now!

objstore.go
```diff
@@ -395,6 +399,101 @@ func DownloadDir(ctx context.Context, logger log.Logger, bkt BucketReader, origi
...
+}
+
+const saltSizeBytes = 32
+
+// As per https://github.com/minio/sio/blob/master/DARE.md#appendices we need a unique key data stream.
+// We derive a unique key from the configuration provided master key by fetching 32 random bits salt and
+// using KDF(master key ++ salt) as our derived encryption key. The salt is prepended to the encrypted
+// object. This is okay since the salt does not need to be kept a secret.
+func (eb *encryptedBucket) encryptionConfig(salt []byte) sio.Config {
+	dk := argon2.Key(eb.masterKey, salt, 3, 32*1024, 4, 32)
```
MichaHoffmann, line 420: this might be too expensive, right?

alanprot:
Idk what block storage you are using but cortex has s3 with sse... maybe is worth looking at it:

https://github.com/cortexproject/cortex/blob/master/pkg/storage/bucket/sse_bucket_client.go

MichaHoffmann:
> Idk what block storage you are using but cortex has s3 with sse... maybe is worth looking at it:
> 
> https://github.com/cortexproject/cortex/blob/master/pkg/storage/bucket/sse_bucket_client.go

I wanted to add client side encryption for situations where server side encryption is not an option though

objstore.go
```diff
@@ -395,6 +399,105 @@ func DownloadDir(ctx context.Context, logger log.Logger, bkt BucketReader, origi
...
+func (eb *encryptedBucket) Get(ctx context.Context, name string) (io.ReadCloser, error) {
+	return eb.GetRange(ctx, name, 0, -1)
+}
+
+func (eb *encryptedBucket) GetRange(ctx context.Context, name string, off, length int64) (io.ReadCloser, error) {
+	saltReader, err := eb.Bucket.GetRange(ctx, name, 0, saltSizeBytes)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to fetch salt")
+	}
+	defer saltReader.Close()
```
GiedriusS, line 470: I believe that we also need to exhaust the reader so that keep-alive connections would be properly maintained.

objstore.go
```diff
@@ -395,6 +399,105 @@ func DownloadDir(ctx context.Context, logger log.Logger, bkt BucketReader, origi
...
+	name string
+	b    BucketReader
+}
+
+func (br *bucketReaderAt) ReadAt(p []byte, off int64) (n int, err error) {
+	rc, err := br.b.GetRange(br.ctx, br.name, off+saltSizeBytes, int64(len(p)))
+	if err != nil {
+		return 0, err
+	}
+	defer rc.Close()
```
GiedriusS, line 496: Same here.

MichaHoffmann: :heavy_check_mark:

objstore.go
MichaHoffmann: :heavy_check_mark:

objstore.go
```diff
@@ -395,6 +399,115 @@ func DownloadDir(ctx context.Context, logger log.Logger, bkt BucketReader, origi
...
+func (eb *encryptedBucket) Get(ctx context.Context, name string) (io.ReadCloser, error) {
+	return eb.GetRange(ctx, name, 0, -1)
+}
+
+func (eb *encryptedBucket) GetRange(ctx context.Context, name string, off, length int64) (io.ReadCloser, error) {
+	metaReader, err := eb.Bucket.GetRange(ctx, name, 0, metaSizeBytes)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to fetch salt")
+	}
+	defer metaReader.Close()
```
GiedriusS, line 477: Let's use: https://github.com/efficientgo/tools/blob/6b7392939a606553331e012253354849d7f9df92/core/pkg/logerrcapture/do.go#L44

PR #50, opened 2023-04-27
Add Huawei Cloud OBS Object Storage Support

setoru:
Huawei Cloud is a public cloud provider in China, Its object storage product is OBS.

thanos is not supported OBS right now,This pr implements a OBS client and it seems to be ok.

I would like to become the maintainer for the OBS provider and keep development for Thanos.

## Changes

- Implement Bucket interface
- Add NewTestBucket constructor
- Pass the ForeachStore test
- Add client implementation to the factory
- Add client struct config to bucketcfggen

## Verification

image


objtesting/foreach.go
```diff
@@ -4,6 +4,7 @@
 package objtesting
 
 import (
+	"github.com/thanos-io/objstore/providers/obs"
```
matej-g, line 7: nit: Can we keep imports grouping? i.e. move this down to the last group.

providers/obs/obs.go
```diff
@@ -0,0 +1,407 @@
...
+	"github.com/prometheus/common/model"
+	"github.com/thanos-io/objstore"
+	"github.com/thanos-io/objstore/exthttp"
+	"gopkg.in/yaml.v2"
+	"io"
+	"math"
+	"os"
+	"strings"
+	"testing"
+	"time"
```
matej-g, line 20: Same as previous comment (let's group imports by std, Thanos and 3rd party)

scripts/cfggen/main.go
```diff
@@ -5,6 +5,7 @@ package main
 
 import (
 	"fmt"
+	"github.com/thanos-io/objstore/providers/obs"
```
matej-g, line 8: Nit: same as previous (import grouping)

client/factory.go
```diff
@@ -6,6 +6,7 @@ package client
 import (
 	"context"
 	"fmt"
+	"github.com/thanos-io/objstore/providers/obs"
```
matej-g, line 9: Nit: Import ordering

providers/obs/obs.go
```diff
@@ -0,0 +1,407 @@
...
+	return nil
+}
+
+func (b *Bucket) putObjectSingle(key string, body io.Reader) error {
+	input := &obs.PutObjectInput{}
+	input.Bucket = b.name
+	input.Key = key
+	input.Body = body
+	_, err := b.client.PutObject(input)
+	return errors.Wrap(err, "fail to upload object")
```
matej-g, line 187: If find doing `if err != nil { return errors.Wrap(...) }` to be a readable a bit better rather than returning directly `errors.Wrap`.

matej-g (commented):
Thanks for this! For now I skimmed through the PR, few suggestions. It also looks like the tests and lint are failing for now.

objtesting/foreach.go
setoru: sure can,I will fix it.

providers/obs/obs.go
setoru: You are right,I will fix it

providers/obs/obs.go
```diff
@@ -0,0 +1,416 @@
...
+		return false, errors.Wrap(err, "fail to get object metadata")
+	}
+	return true, nil
+}
+
+// IsObjNotFoundErr returns true if error means that object is not found. Relevant to Get operations.
+func (b *Bucket) IsObjNotFoundErr(err error) bool {
+	switch oriErr := errors.Cause(err).(type) {
+	case obs.ObsError:
+		if oriErr.Status == "404 Not Found" {
```
matej-g, line 325: Since we have only one case, why not do `oriErr, ok := errors.Cause(err).(obs.ObsError)`?

providers/obs/obs.go
```diff
@@ -248,16 +248,16 @@ func (b *Bucket) Iter(ctx context.Context, dir string, f func(string) error, opt
...
 		for _, content := range output.Contents {
 			if err := f(content.Key); err != nil {
-				return errors.Wrap(err, "fail to call f for object")
+				return errors.Wrap(err, "failed to call iter function for object X")
 			}
 		}
 		for _, topDir := range output.CommonPrefixes {
 			if err := f(topDir); err != nil {
-				return errors.Wrap(err, "fail to call f for top dir object")
+				return errors.Wrap(err, "failed to call f for top dir object")
```
matej-g, line 260: Can we adjust this the same way as my previous comment (don't say `failed to call f ...` but rather `failed to call iter function`)

matej-g (approved):
This is looking good now, thank you @setoru! 🙇

setoru:
Thank you for reviewing!

PR #51, opened 2023-04-28
azure: Support using connection string authentication

nholik:

## Changes

Added an additional option `storage_connection_string` to Azure config to use a connection string rather than a key.  This is needed in limited privilege scenarios where only a SAS token is available and not  full access key.  When provided, the value is forwarded on to the Azure SDK method for creating a client that accepts a connection string.  

Existing consumers will be unaffected as using this new path must be opted into with a non-empty value for the new config option.

## Verification

Using a connection string with a SAS token creates a client successfully.  Further verified that running a patched version of loki consuming this library passing through SAS tokens as the authentication method works.

### DIFF
--- a/providers/azure/azure_test.go
+++ b/providers/azure/azure_test.go
@@ -132,6 +132,16 @@ container: "MyContainer"`),
 		wantFailParse:    false,
 		wantFailValidate: false,
 	},
+	{
+		name: "Valid User Assigned and Connection String set",
+		config: []byte(`storage_account: "myAccount"
+storage_account_key: ""
+user_assigned_id: "1234-56578678-655"
+storage_connection_string: "myConnectionString"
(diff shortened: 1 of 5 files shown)


nholik:
> LGTM.
> 
> Needs a rebase, though.
Thanks @kakkoyun - rebase done

providers/azure/azure_test.go
```diff
@@ -132,6 +132,16 @@ container: "MyContainer"`),
 		wantFailParse:    false,
 		wantFailValidate: false,
 	},
+	{
+		name: "Valid User Assigned and Connection String set",
+		config: []byte(`storage_account: "myAccount"
+storage_account_key: ""
```
christopherzli, line 138: how is this valid  since storage_account_key and storage_connection_string cannot both be set?

PR #59, opened 2023-06-23
Implementing check to identify CMK errors (s3)

alanprot:
Some bucket like [s3](https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingKMSEncryption.html) and [Azure](https://learn.microsoft.com/en-us/azure/storage/common/customer-managed-keys-overview) implementations allows encrypting the block storage with customer managed keys and the permissions to this keys can be revoked anytime.

This PR is creating a extra method (`IsCustomerManagedKeyError`) when the key permission got revoked by the key owner and so the upstream can handle it appropriately (cortex,thanos)

Related to: https://github.com/cortexproject/cortex/pull/5420

As a first step only the s3 implementation is identifying such errors but i can follow up with the azure one.

* [ X] I added CHANGELOG entry for this change.

## Changes

* Add `IsCustomerManagedKeyError` method on the bucket interface.
* Implement `IsCustomerManagedKeyError` on the s3 bucket implementation.

### DIFF
--- a/objstore.go
+++ b/objstore.go
@@ -85,6 +85,9 @@ type BucketReader interface {
 	// IsObjNotFoundErr returns true if error means that object is not found. Relevant to Get operations.
 	IsObjNotFoundErr(err error) bool
 
+	// IsCustomerManagedKeyError returns true if the permissions for key used to encrypt the object was revoked.
+	IsCustomerManagedKeyError(err error) bool
+
 	// Attributes returns information about the specified object.
 	Attributes(ctx context.Context, name string) (ObjectAttributes, error)
 }
@@ -601,6 +604,10 @@ func (b *metricBucket) Delete(ctx context.Context, name string) error {
 
 func (b *metricBucket) IsObjNotFoundErr(err error) bool {
 	return b.bkt.IsObjNotFoundErr(err)
+}
+
+func (b *metricBucket) IsCustomerManagedKeyError(err error) bool {
+	return b.bkt.IsCustomerManagedKeyError(err)
 }
 
 func (b *metricBucket) Close() error {
--- a/providers/s3/s3.go
+++ b/providers/s3/s3.go
@@ -98,6 +98,9 @@ const (
 
 	// Storage class header.
 	amzStorageClass = "X-Amz-Storage-Class"
+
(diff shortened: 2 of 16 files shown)


alanprot:
@GiedriusS @fpetkovski Thoughts?

I wanna use to identify when the key access was revoked so we can return an well defined error from store gateway: on cortex: https://github.com/cortexproject/cortex/pull/5420

objstore.go
```diff
@@ -85,6 +85,9 @@ type BucketReader interface {
 	// IsObjNotFoundErr returns true if error means that object is not found. Relevant to Get operations.
 	IsObjNotFoundErr(err error) bool
 
+	// IsCustomerManagedKeyError returns true if the permissions for key used to encrypt the object was revoked.
+	IsCustomerManagedKeyError(err error) bool
```
fpetkovski, line 89: Could we return a newly defined `ErrInvalidCustomerManagedKey` error and use `errors.Cause` to identify it? This way a client does not need an instance of the bucket in order to distinguish between error types.

fpetkovski: I missed that we already have `IsObjNotFoundErr`, so I guess this should also be fine.

fpetkovski (approved):
Looks good to me, but would be nice for someone with more knowledge of the codebase to take a look as well.

providers/s3/s3.go
```diff
@@ -538,6 +538,12 @@ func (b *Bucket) IsObjNotFoundErr(err error) bool {
 	return minio.ToErrorResponse(errors.Cause(err)).Code == "NoSuchKey"
 }
 
+// IsCustomerManagedKeyError returns true if the permissions for key used to encrypt the object was revoked.
+func (b *Bucket) IsCustomerManagedKeyError(err error) bool {
+	errResponse := minio.ToErrorResponse(errors.Cause(err))
+	return errResponse.Code == "AccessDenied" && errResponse.Message == "The ciphertext refers to a customer master key that does not exist, does not exist in this region, or you are not allowed to access."
```
yeya24, line 544: Regarding this error message, is it possible for it to change? Or can we import this somewhere from other libraries like Minio?
Or we can make it a constant first.

PR #61, opened 2023-07-06
Add OTel tracing bucket

kakkoyun:
- Refactor client.NewBucket. This changes the behaviour of `client.NewBucket`. Now it returns, uninstrumented and untraced bucket.
- Add OpenTelemetry TracingBucket.

Now you can combine `objstore.WrapWithMetrics` and `tracing/{opentelemetry,opentracing}.WrapWithTraces` to have old behavior.

Signed-off-by: Kemal Akkoyun <kakkoyun@gmail.com>

## Changes

- Add OTel Tracing bucket
- Introduce tracing package
- Refactor client.NewBucket
- Introduce client.NewInstrumentedBucket
- Add examples


client/factory.go
```diff
@@ -91,5 +91,11 @@ func NewBucket(logger log.Logger, confContentYaml []byte, reg prometheus.Registe
 		return nil, errors.Wrap(err, fmt.Sprintf("create %s client", bucketConf.Type))
 	}
 
-	return objstore.NewTracingBucket(objstore.BucketWithMetrics(bucket.Name(), objstore.NewPrefixedBucket(bucket, bucketConf.Prefix), reg)), nil
+	return objstore.NewPrefixedBucket(bucket, bucketConf.Prefix), nil
+}
+
+// NewInstrumentedBucket initializes and returns new object storage clients.
+// NOTE: confContentYaml can contain secrets.
```
brancz, line 98: this note is not relevant here

README.md
```diff
@@ -643,18 +644,19 @@ You can also include any of the optional configuration just like the example in
...
 
 To configure HuaweiCloud Account to use OBS as storage store you need to set these parameters in YAML format stored in a file:
 
 ```yaml mdox-exec="go run scripts/cfggen/main.go --name=cos.Config"
-type: OBS                    
-config:                      
-  bucket: ""                 
-  endpoint: ""               
-  access_key: ""             
+type: COS
```
brancz, line 652: is this correct? It talks about Huawei OBS above and then COS?

brancz (commented):
I'm happy with this. Are you going to take care of fixing Thanos to make sure it doesn't loose instrumentation?

README.md
kakkoyun: Let me double-check. I think running the linter or some other make action caused this.

kakkoyun:
> I'm happy with this. Are you going to take care of fixing Thanos to make sure it doesn't lose instrumentation?

I'll make sure to do that. Maybe we can use opentracing tracing for the instrumented bucket? WDYT?

README.md
kakkoyun: Thanks for catching it. I've fixed the command above that generates this doc.

brancz:
I think just replacing the spots where Thanos just called NewBucket, after this merges replace that with `NewTracingBucket(NewInstrumentedBucket(NewBucket(...)))`. That would be sufficient.

kakkoyun:
> I think just replacing the spots where Thanos just called NewBucket, after this merges replace that with `NewTracingBucket(NewInstrumentedBucket(NewBucket(...)))`. That would be sufficient.

https://github.com/thanos-io/thanos/pull/6507

tracing/tracing.go
```diff
@@ -34,36 +28,11 @@ func ContextWithTracer(ctx context.Context, tracer opentracing.Tracer) context.C
 	return context.WithValue(ctx, tracerKey, tracer)
 }
 
-// tracerFromContext extracts opentracing.Tracer from the given context.
-func tracerFromContext(ctx context.Context) opentracing.Tracer {
+// TracerFromContext extracts opentracing.Tracer from the given context.
+func TracerFromContext(ctx context.Context) opentracing.Tracer {
```
bwplotka, line 32: Do we need this? Can we move it to opentracing, so "common" tracing pkg is vendor free?

client/factory.go
```diff
@@ -91,5 +91,10 @@ func NewBucket(logger log.Logger, confContentYaml []byte, reg prometheus.Registe
 		return nil, errors.Wrap(err, fmt.Sprintf("create %s client", bucketConf.Type))
 	}
 
-	return objstore.NewTracingBucket(objstore.BucketWithMetrics(bucket.Name(), objstore.NewPrefixedBucket(bucket, bucketConf.Prefix), reg)), nil
+	return objstore.NewPrefixedBucket(bucket, bucketConf.Prefix), nil
```
bwplotka, line 94: Hm, I think it's worth to mention in change log AND in comment of NewBucket, that this NewBucket no longer instruments and adds span middleware. Callers have to add them on their own now.

client/factory.go
```diff
@@ -91,5 +91,10 @@ func NewBucket(logger log.Logger, confContentYaml []byte, reg prometheus.Registe
...
+}
+
+// NewInstrumentedBucket initializes and returns new object storage clients.
```
bwplotka, line 97: This is not true - it only wraps with metrics. Should we rather remove `NewInstrumentedBucket` and keep `objstore.BucketWithMetrics`?

client/factory.go
```diff
@@ -91,5 +91,10 @@ func NewBucket(logger log.Logger, confContentYaml []byte, reg prometheus.Registe
...
+func NewInstrumentedBucket(reg prometheus.Registerer, bkt objstore.Bucket) objstore.InstrumentedBucket {
+	return objstore.BucketWithMetrics(bkt.Name(), bkt, reg)
```
bwplotka, line 99: Renamed this:
```suggestion
objstore.InstrumentedBucket(bkt.Name(), bkt, reg)
```

so we have "InstrumentedBucket", "otel.TracingBucket", "opentracing.TracingBucket"? 🤔

client/testconf/fake-gcs.conf.yml
```diff
@@ -1,3 +0,0 @@
-type: FAKE-GCS
```
bwplotka, line 1: Why changing, sounds like we remve some test cases which right now are uncovered? 🤔

tracing/opentelemetry/opentelemetry.go
```diff
@@ -0,0 +1,190 @@
...
+	"github.com/thanos-io/objstore"
+)
+
+// TracingBucket is a wrapper around objstore.Bucket that adds tracing to all operations using OpenTelemetry.
+type TracingBucket struct {
+	tracer trace.Tracer
+	bkt    objstore.Bucket
+}
+
+func NewTracingBucket(tracer trace.Tracer, bkt objstore.Bucket) objstore.InstrumentedBucket {
```
bwplotka, line 22: While we are here we could stop calling it "new" 🤔 . It's not new bucket. It's not even new bucket provider. Perhaps  `otel.TracedBucket(...)`or `WrapWithTracing`

bwplotka (commented):
Nice! Some comments, otherwise LGTM 👍🏽

client/testconf/fake-gcs.conf.yml
kakkoyun: These were not utilized. I haven't tracked them down but my they are left overs from the migration.

brancz (approved):
lgtm

client/factory.go
```diff
@@ -91,5 +91,10 @@ func NewBucket(logger log.Logger, confContentYaml []byte, reg prometheus.Registe
 		return nil, errors.Wrap(err, fmt.Sprintf("create %s client", bucketConf.Type))
 	}
 
-	return objstore.NewTracingBucket(objstore.BucketWithMetrics(bucket.Name(), objstore.NewPrefixedBucket(bucket, bucketConf.Prefix), reg)), nil
+	return objstore.NewPrefixedBucket(bucket, bucketConf.Prefix), nil
+}
+
+// InstrumentedBucket wraps the given bucket with metrics.
+func InstrumentedBucket(bkt objstore.Bucket, reg prometheus.Registerer) objstore.InstrumentedBucket {
```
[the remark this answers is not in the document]
kakkoyun: I'm ok with this approach as well. 
If @bwplotka has nothing else to fix, I'll remove it and merge the PR.

kakkoyun:
Ping @bwplotka, is it ok to merge now?

client/factory.go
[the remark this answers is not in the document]
bwplotka: Bit confused, I suggested removing this in https://github.com/thanos-io/objstore/pull/61#discussion_r1254437185 and sounds like @brancz also suggest removal of this function. Let's address this before merging (either removing or letting us know why removal does not make sense 🙃

client/testconf/fake-gcs.conf.yml
bwplotka: 👍🏽

tracing/opentelemetry/opentelemetry.go
```diff
@@ -0,0 +1,190 @@
+// Copyright (c) The Thanos Authors.
+// Licensed under the Apache License 2.0.
+
+package opentelemetry
```
bwplotka, line 4: Can be simplified to `otel` if we want:

```suggestion
package otel
```

bwplotka (approved):
Thanks! Still some unresolved comments, otherwise good to go!

tracing/opentelemetry/opentelemetry.go
kakkoyun: Let's keep naming consistent with already existing parts (`opentracing` and `opentelemetry`).

bwplotka (approved):

tracing/opentelemetry/opentelemetry.go
```diff
@@ -0,0 +1,190 @@
...
+	defer func() {
+		if err != nil {
+			span.RecordError(err)
+		}
+	}()
+	return t.bkt.Delete(ctx, name)
+}
+
+func (t TracingBucket) Name() string {
+	return "tracing: " + t.bkt.Name()
```
bboreham, line 120: Why add this?

PR #62, opened 2023-07-06
S3: Return context cancellation from Iter when applicable

56quarters:
## Changes

This PR returns `ctx.Err()` from the S3 `Iter` method instead of `nil` after exhausting the `ListObjects` channel.

When the context passed to the S3 Iter method is cancelled, the underlying Minio client sometimes immediately [closes the results channel](https://github.com/grafana/mimir/blob/285fa4e35dd5720e38dd0491f6ddeb51e3e81197/vendor/github.com/minio/minio-go/v7/api-list.go#L100) and returns without sending the `ctx.Err()` error. Depending on [when exactly the context is cancelled](https://github.com/grafana/mimir/blob/285fa4e35dd5720e38dd0491f6ddeb51e3e81197/vendor/github.com/minio/minio-go/v7/api-list.go#L114-L136), this can result in the `Iter` method not returning any results _and_ not returning an error indicating what happened.

Discovered while debugging grafana/mimir#5433

## Verification

Tested via [Mimir end-to-end test used to reproduce the issue](https://github.com/grafana/mimir/pull/5427). The S3 `Bucket` implementation uses the concrete Minio `Client` instead of an interface so writing a unit test would be a challenge without making quite a few changes to the `Bucket`. I can do that if you'd like but it seemed like a lot of churn for a small change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 ### Fixed
 - [#33](https://github.com/thanos-io/objstore/pull/33) Tracing: Add `ContextWithTracer()` to inject the tracer into the context.
 - [#34](https://github.com/thanos-io/objstore/pull/34) Fix ignored options when creating shared credential Azure client.
+- [#62](https://github.com/thanos-io/objstore/pull/62) S3: Fix ignored context cancellation in `Iter` method. 
 
 ### Added
 - [#15](https://github.com/thanos-io/objstore/pull/15) Add Oracle Cloud Infrastructure Object Storage Bucket support.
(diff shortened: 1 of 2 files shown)


GiedriusS (approved):
LGTM, however, could you please also create an upstream issue in minio-go? It seems like this is something that could be solved there. I also saw this in prod where remote object storage was having problems and sometimes the meta fetcher returned 0 metas with no errors. This is probably the cause. Haven't had time to dig into this myself.

56quarters:
> LGTM, however, could you please also create an upstream issue in minio-go? It seems like this is something that could be solved there. I also saw this in prod where remote object storage was having problems and sometimes the meta fetcher returned 0 metas with no errors. This is probably the cause. Haven't had time to dig into this myself.

Opened an [issue](https://github.com/minio/minio-go/issues/1850), thanks!

PR #63, opened 2023-07-14
Add support for object attributes in Iter call

fpetkovski:
Add support for IterWithAttributes

This commit adds support for an IterWithAttributes on the bucket client.
The method allows iterating through objects and getting multiple attributes
into the callback function, removing the need to do an Iter followed by Attrs.

For now, we only support getting the last updated time as an attribute, but the
implementation allows adding more in the future.

Not all buckets support this method. The client can check whether the bucket has
support by calling the SupportedIterOptions method on the client.

Co-authored-by: Ashwanth Goli <iamashwanth@gmail.com>
Co-authored-by: Filip Petkovski <filip.petkovsky@gmail.com>

Signed-off-by: Filip Petkovski <filip.petkovsky@gmail.com>

## Changes

* Add support for `IterWithAttributes`. 

## Verification

* Unit tests.

### DIFF
--- a/objstore.go
+++ b/objstore.go
@@ -604,14 +604,15 @@ func (b *metricBucket) Iter(ctx context.Context, dir string, f func(string) erro
 	const op = OpIter
 	b.metrics.ops.WithLabelValues(op).Inc()
 
-	start := time.Now()
+	timer := prometheus.NewTimer(b.metrics.opsDuration.WithLabelValues(op))
+	defer timer.ObserveDuration()
+
 	err := b.bkt.Iter(ctx, dir, f, options...)
 	if err != nil {
 		if !b.metrics.isOpFailureExpected(err) && ctx.Err() != context.Canceled {
 			b.metrics.opsFailures.WithLabelValues(op).Inc()
 		}
 	}
-	b.metrics.opsDuration.WithLabelValues(op).Observe(time.Since(start).Seconds())
 	return err
 }
 
@@ -619,12 +620,16 @@ func (b *metricBucket) IterWithAttributes(ctx context.Context, dir string, f fun
 	const op = OpIter
 	b.metrics.ops.WithLabelValues(op).Inc()
 
+	timer := prometheus.NewTimer(b.metrics.opsDuration.WithLabelValues(op))
+	defer timer.ObserveDuration()
+
 	err := b.bkt.IterWithAttributes(ctx, dir, f, options...)
 	if err != nil {
 		if !b.metrics.isOpFailureExpected(err) && ctx.Err() != context.Canceled {
 			b.metrics.opsFailures.WithLabelValues(op).Inc()
 		}
 	}
+
 	return err
 }
 
--- a/objstore.go
+++ b/objstore.go
@@ -6,11 +6,13 @@ package objstore
 import (
 	"bytes"
 	"context"
+	"fmt"
 	"io"
 	"io/fs"
 	"os"
 	"path"
 	"path/filepath"
+	"slices"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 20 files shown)


prefixed_bucket.go
```diff
@@ -46,11 +46,11 @@ func (p *PrefixedBucket) Close() error {
...
 // object name including the prefix of the inspected directory.
 // Entries are passed to function in sorted order.
-func (p *PrefixedBucket) Iter(ctx context.Context, dir string, f func(string) error, options ...IterOption) error {
+func (p *PrefixedBucket) Iter(ctx context.Context, dir string, f func(name string, attrs ObjectAttributes) error, options ...IterOption) error {
 	pdir := withPrefix(p.prefix, dir)
 
-	return p.bkt.Iter(ctx, pdir, func(s string) error {
-		return f(strings.TrimPrefix(s, p.prefix+DirDelim))
+	return p.bkt.Iter(ctx, pdir, func(s string, _ ObjectAttributes) error {
+		return f(strings.TrimPrefix(s, p.prefix+DirDelim), EmptyObjectAttributes)
```
brancz, line 53: why wouldn't it be passed through here?

objstore.go
```diff
@@ -68,10 +70,12 @@ type InstrumentedBucket interface {
...
 type BucketReader interface {
-	// Iter calls f for each entry in the given directory (not recursive.). The argument to f is the full
-	// object name including the prefix of the inspected directory.
+	// Iter calls f for each entry in the given directory (not recursive.). The first argument to f is the full
+	// object name including the prefix of the inspected directory. The second argument are the object attributes
+	// returned by the underlying provider. Attributes can be requested using various IterOption's.
+	//
 	// Entries are passed to function in sorted order.
-	Iter(ctx context.Context, dir string, f func(string) error, options ...IterOption) error
+	Iter(ctx context.Context, dir string, f func(name string, attrs ObjectAttributes) error, options ...IterOption) error
```
brancz, line 78: If we're already breaking this signature, I think we should instead make the whole thing pass _only_ the `ObjectAttributes` struct and add the `Name` field to it.

brancz (commented):
I like this change a lot!

objstore.go
fpetkovski: This makes sense, and would allow us to extend attributes in the future.

brancz: Exactly!

fpetkovski:
Yeah I will look into adding a new APi. I am still not sure how to handle cases where a provider does not support returning attributes.

pracucci:
> Yeah I will look into adding a new APi. I am still not sure how to handle cases where a provider does not support returning attributes.

An idea may be having a function on the objstore client to check if the feature is supported. Then the Iter() contract is that if the feature is supported, attributes will be fetched, otherwise if the feature is not supported it will error out. It's caller responsability to check if the feature is supported by the actual objstore client and take an informed decision about requesting attributes or not.

ashwanthgoli:
@fpetkovski are there plans to revive this pr? :)

Loki project is planning to switch to thanos clients. Having this change will reduce the fanout of listing objects along with metadata which now requires Iter call followed by fetching attributes for all objects

objstore.go
```diff
@@ -101,24 +114,66 @@ type InstrumentedBucketReader interface {
...
+		Apply: func(params *IterParams) {
+			params.Recursive = true
+		},
+	}
+}
+
+// WithUpdatedAt is an option that can be applied to Iter() to
+// include the last modified time in the attributes.
+// NB: Prefixes may not report last modified time.
+// This option is currently supported for the azure, aws, bos, gcs and filesystem providers.
```
JoaoBraveCoding, line 147: nit: just to match the provider name
```suggestion
// This option is currently supported for the azure, s3, bos, gcs and filesystem providers.
```

objstore.go
```diff
@@ -189,6 +244,20 @@ type ObjectAttributes struct {
...
 }
 
+type IterObjectAttributes struct {
+	Name         string
+	lastModified time.Time
+}
+
+func (i *IterObjectAttributes) SetLastModified(t time.Time) {
+	i.lastModified = t
+}
```
JoaoBraveCoding, line 254: Curious question: why did we make assignments to lastModified come through a setter but not name?

JoaoBraveCoding (approved):
as far as I can see lgtm

providers/gcs/gcs.go
```diff
@@ -181,31 +181,33 @@ func (b *Bucket) Name() string {
...
+
 	// If recursive iteration is enabled we should pass an empty delimiter.
 	delimiter := DirDelim
-	if objstore.ApplyIterOptions(options...).Recursive {
+	if appliedOpts.Recursive {
 		delimiter = ""
 	}
 
-	query := &storage.Query{
+	it := b.bkt.Objects(ctx, &storage.Query{
```
ashwanthgoli, line 207: @fpetkovski i didn't realise this was only fetching `Name` attribute before.
I think I removed changes from a previous commit of yours, we need to add it back :)

objstore.go
```diff
@@ -546,6 +615,23 @@ func (b *metricBucket) Iter(ctx context.Context, dir string, f func(name string)
 	return err
 }
 
+func (b *metricBucket) IterWithAttributes(ctx context.Context, dir string, f func(IterObjectAttributes) error, options ...IterOption) error {
+	const op = OpIter
+	b.metrics.ops.WithLabelValues(op).Inc()
+
+	err := b.bkt.IterWithAttributes(ctx, dir, f, options...)
```
ashwanthgoli, line 622: updates to `b.metrics.opsDuration` metric is missing

objstore.go
ashwanthgoli: i wanted to make it explicit that `lastModified` might not always be set, made it a private field so the getter contract can return `time.Time, bool`

ashwanthgoli:
thank you! @fpetkovski
i left a couple of comments, but the rest lgtm

objstore.go
ashwanthgoli: LGTM!

PR #66, opened 2023-07-21
objstore: Add uploaded TSDB bytes metric

ritaCanavarro:

## Changes

As suggested on https://github.com/thanos-io/thanos/pull/6544#pullrequestreview-1541053296 the metric regarding the number of bytes the shipper has uploaded is best suited to be on the objstore. Therefore, this PR aims to add it to the objstore.go class.

## Verification

I updated the TestTimingTracingReader test in objstore_test.go but I can add more tests if needed.

Thanks!


CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#61](https://github.com/thanos-io/objstore/pull/61) Add OpenTelemetry TracingBucket.
     > This also changes the behaviour of `client.NewBucket`. Now it returns, uninstrumented and untraced bucket.
     You can combine `objstore.WrapWithMetrics` and `tracing/{opentelemetry,opentracing}.WrapWithTraces` to have old behavior.
+- [#66](https://github.com/thanos-io/objstore/pull/66) Add for the number of total uploaded blocks.
```
bwplotka, line 30:
```suggestion
- [#66](https://github.com/thanos-io/objstore/pull/66) Add `objstore_bucket_operation_written_bytes_total` that counts the number of total bytes per written to the bucket operation.
```

objstore.go
```diff
@@ -435,6 +435,12 @@ func WrapWithMetrics(b Bucket, reg prometheus.Registerer, name string) *metricBu
 			Name: "objstore_bucket_last_successful_upload_time",
 			Help: "Second timestamp of the last successful upload to the bucket.",
 		}, []string{"bucket"}),
+
+		opsUploadedBytes: promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
+			Name:        "objstore_bucket_operation_uploaded_bytes_total",
```
bwplotka, line 440:
```suggestion
			Name:        "objstore_bucket_operation_written_bytes_total",
```

We have two options. Either objstore_bucket_operation_written_bytes_total and make sure we increment every time we write (upload) some bytes to bucket. Technically the only significant bytes are written in Upload method, but we don't know in future, so let's keep operation label and put "write" in name so we are flexible here.

objstore.go
```diff
@@ -631,9 +642,10 @@ type timingReadCloser struct {
 	failed            *prometheus.CounterVec
 	isFailureExpected IsOpFailureExpectedFunc
 	fetchedBytes      *prometheus.CounterVec
+	uploadedBytes     *prometheus.CounterVec
```
bwplotka, line 645: I don't think we increment it anywhere? Did you miss incrementing it?

We also really need this metric for Upload operation. No other operation writes significant number of bytes unless I am wrong (:

bwplotka (commented):
Epic! Good start, some suggestions for the start (: 

Approved CI, will try to fix setting so it allows you to use CI without approvals

objstore.go
ritaCanavarro: I missed it too, my bad.
I have added it now.

simonswine:
I was separately working in exposing the fetched bytes as a histogram. Wonder if it makes sense to integrate both of our work into a single histogram: https://github.com/thanos-io/objstore/pull/69

ritaCanavarro:
> I was separately working in exposing the fetched bytes as a histogram. Wonder if it makes sense to integrate both of our work into a single histogram: #69

Hi  @simonswine, if I understood correctly, it seems that we are working on different subjects so perhaps we should keep our PRs separated.

bwplotka:
Merged your PR @simonswine.

I think @simonswine has a good point. We have two options: 

A. Since we have both total fetched and distribution of fetched bytes per op, we could have similar for written bytes (total and later histogram in separate PR)
B.  Although histogram's _sum might gives the same info as total counter, so we could skip adding upload bytes and focus in this PR to only add histogram (this should be easy @ritaCanavarro - just change type to histogram and use "Observe" instead of increments).

I don't have strong opinion,  I trust @simonswine you have good judgment here. Would you like to help @ritaCanavarro to get this contribution to mergable state? 🤗

bwplotka:
Also, @ritaCanavarro you need to rebase now (:

simonswine:
I am also okay with both options. 

And of course happy to help out @ritaCanavarro with this contribution.

objstore.go
```diff
@@ -442,6 +442,13 @@ func WrapWithMetrics(b Bucket, reg prometheus.Registerer, name string) *metricBu
 			Name: "objstore_bucket_last_successful_upload_time",
 			Help: "Second timestamp of the last successful upload to the bucket.",
 		}, []string{"bucket"}),
+
+		opsWrittenBytes: promauto.With(reg).NewHistogramVec(prometheus.HistogramOpts{
+			Name:        "objstore_bucket_operation_written_bytes",
+			Help:        "Number of bytes uploaded from TSDB block, per operation.",
+			ConstLabels: prometheus.Labels{"bucket": name},
+			Buckets:     prometheus.ExponentialBuckets(50<<100, 2, 10),
+		}, []string{"operation"}),
```
simonswine, line 451: I think it is better rather than using a separate metric, use the transferred bytes histogram vec metric, as it is named so that it allows to measure up and downloads and then the label `operation` would determine if it is in fact an upload or download.

objstore.go
```diff
@@ -593,12 +603,23 @@ func (b *metricBucket) Upload(ctx context.Context, name string, r io.Reader) err
 	b.ops.WithLabelValues(op).Inc()
 
 	start := time.Now()
-	if err := b.bkt.Upload(ctx, name, r); err != nil {
+	err := b.bkt.Upload(ctx, name, r)
```
simonswine, line 606: To be able to count the transferred bytes, you would need to wrap `r`, so you are able get the bytes transferred.

I think you can reuse `newTimingReaderCloser` to wrap this reader here.

objstore.go
ritaCanavarro: Removed the metric and will work on the remaining comments.

README.md
```diff
@@ -57,7 +57,7 @@ type Bucket interface {
 
 	// Upload the contents of the reader as an object into the bucket.
 	// Upload should be idempotent.
-	Upload(ctx context.Context, name string, r io.Reader) error
```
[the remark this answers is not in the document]
ritaCanavarro: As discussed, we will use the reader to get the correct number of uploaded bytes.

objstore.go
ritaCanavarro: Agree and will implement it.

objstore.go
```diff
@@ -593,12 +603,23 @@ func (b *metricBucket) Upload(ctx context.Context, name string, r io.Reader) err
...
 		}
 		return err
 	}
+
+	rc := &nopCloserWithObjectSize{
+		Reader: r,
+	}
+
+	rc.Close()
+	if writtenBytes, err := rc.ObjectSize(); err != nil {
```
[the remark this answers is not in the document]
ritaCanavarro: How about we fetch the object size only after the upload? This way we will only perform this operation if the upload was indeed successful, sparing us from having to compute the size without needing it.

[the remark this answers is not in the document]
simonswine: I agree with that part that it should only count successful uploads with the histogram, but I don't think we need to use a method like `.ObjectSize()`. 

Let look at an example, how the method would be used:

```
// program uploads something
reader := bytes.NewReader([]byte("my-file-content")
Upload(ctx, "my-file.txt", reader)
```

Then it would end up in the `metricsBucket.Upload` which in turn calls `provider.Upload`.

Then in the provider implementation it would retrieve the content by calling `reader.Read()` at least once, but potentially multiple times.

This `Read()` method returns each time `n int,err error`, and `n` is the amount of bytes read, from the reader. So to get the uploaded size we need to sum up the `n`s returned by `Read` which will be the exact amount of bytes that have been uploaded.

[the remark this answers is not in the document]
ritaCanavarro: Sorry I was not clear on my part, I used the words object size but I did not mean the Object size function but the size of the object uploaded 😊 

Will apply these changes as discussed.

objstore.go
ritaCanavarro: @simonswine can you please take a look?

simonswine: @ritaCanavarro I think we are getting pretty close, the main concern is that the uploaded bytes do not get recorded yet.

I would expect `objstore_bucket_operation_transferred_bytes_sum{bucket="",operation="upload"} 0` to be `> 0`.


If you look at the line of code above you pass on the unwrapped reader to `r` unmodified to `b.bkt.Upload`, instead you'll need to pass the wrapped reader `trc` to `Upload`, so you are able to count the bytes as they get uploaded.

ritaCanavarro: Thanks @simonswine, I fixed it.

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#61](https://github.com/thanos-io/objstore/pull/61) Add OpenTelemetry TracingBucket.
     > This also changes the behaviour of `client.NewBucket`. Now it returns, uninstrumented and untraced bucket.
     You can combine `objstore.WrapWithMetrics` and `tracing/{opentelemetry,opentracing}.WrapWithTraces` to have old behavior.
+- [#66](https://github.com/thanos-io/objstore/pull/66) Add `objstore_bucket_operation_written_bytes_total` that counts the number of total bytes per written to the bucket operation.
```
simonswine, line 30: This needs an update, also should mention the sizes for Get/GetRange (missed that in my PR #69)

objstore.go
```diff
@@ -593,14 +593,29 @@ func (b *metricBucket) Upload(ctx context.Context, name string, r io.Reader) err
 	b.ops.WithLabelValues(op).Inc()
 
 	start := time.Now()
-	if err := b.bkt.Upload(ctx, name, r); err != nil {
+	trc := newTimingReadCloser(
+		nopCloserWithObjectSize{
+			Reader: r,
+		},
```
simonswine, line 599: Nit: Using `io.NopCloser(r)` would be more appropriate

objstore.go
```diff
@@ -593,14 +593,29 @@ func (b *metricBucket) Upload(ctx context.Context, name string, r io.Reader) err
...
+	if err != nil {
 		if !b.isOpFailureExpected(err) && ctx.Err() != context.Canceled {
 			b.opsFailures.WithLabelValues(op).Inc()
 		}
 		return err
 	}
 	b.lastSuccessfulUploadTime.SetToCurrentTime()
 	b.opsDuration.WithLabelValues(op).Observe(time.Since(start).Seconds())
+
+	trc.Close()
```
simonswine, line 617: When there is an error in `607`, we won't call `trc.Close()`, we should also call it then.

simonswine (commented):
Thank you @ritaCanavarro for putting the hard work in. I think we have it working, just a few more cleanup as per mit comments

objstore_test.go
```diff
@@ -126,7 +126,7 @@ func TestDownloadUploadDirConcurrency(t *testing.T) {
         objstore_bucket_operation_fetched_bytes_total{bucket="",operation="get"} 1.048578e+06
         objstore_bucket_operation_fetched_bytes_total{bucket="",operation="get_range"} 0
         objstore_bucket_operation_fetched_bytes_total{bucket="",operation="iter"} 0
-        objstore_bucket_operation_fetched_bytes_total{bucket="",operation="upload"} 0
+        objstore_bucket_operation_fetched_bytes_total{bucket="",operation="upload"} 1.048578e+06
```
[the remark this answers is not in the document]
ritaCanavarro: Should I do what you are doing in this PR? https://github.com/thanos-io/objstore/pull/68/files

CHANGELOG.md
ritaCanavarro: Added a new message for your PR and mine, if you could check?

objstore.go
```diff
@@ -593,14 +593,29 @@ func (b *metricBucket) Upload(ctx context.Context, name string, r io.Reader) err
...
 
 	start := time.Now()
-	if err := b.bkt.Upload(ctx, name, r); err != nil {
+	trc := newTimingReadCloser(
+		io.NopCloser(r),
+		op,
+		b.opsDuration,
+		b.opsFailures,
+		b.isOpFailureExpected,
+		b.opsFetchedBytes,
```
simonswine, line 602: I would use `nil` here and then check to avoid nil pointer execptions

objstore_test.go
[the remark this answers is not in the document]
simonswine: I don't think that will work, instead I would do this: https://github.com/thanos-io/objstore/pull/66/files#r1287319813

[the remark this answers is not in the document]
ritaCanavarro: Added, please review.

CHANGELOG.md
ritaCanavarro: Added, please review.

objstore_test.go
```diff
@@ -170,6 +170,25 @@ func TestDownloadUploadDirConcurrency(t *testing.T) {
         objstore_bucket_operation_transferred_bytes_bucket{bucket="",operation="get_range",le="+Inf"} 0
         objstore_bucket_operation_transferred_bytes_sum{bucket="",operation="get_range"} 0
         objstore_bucket_operation_transferred_bytes_count{bucket="",operation="get_range"} 0
+		objstore_bucket_operation_transferred_bytes_bucket{bucket="",operation="upload",le="32768"} 2
+        objstore_bucket_operation_transferred_bytes_bucket{bucket="",operation="upload",le="65536"} 2
```
simonswine, line 174: Nit: Whitespace problem

```suggestion
        objstore_bucket_operation_transferred_bytes_bucket{bucket="",operation="upload",le="32768"} 2
        objstore_bucket_operation_transferred_bytes_bucket{bucket="",operation="upload",le="65536"} 2
```

objstore.go
```diff
@@ -593,14 +594,29 @@ func (b *metricBucket) Upload(ctx context.Context, name string, r io.Reader) err
...
 			b.opsFailures.WithLabelValues(op).Inc()
 		}
+
+		trc.Close()
 		return err
 	}
 	b.lastSuccessfulUploadTime.SetToCurrentTime()
 	b.opsDuration.WithLabelValues(op).Observe(time.Since(start).Seconds())
+
+	trc.Close()
```
simonswine, line 618: A bit more fail safe (probably also elegant) would be the use of `defer` here

```suggestion
	trc := newTimingReadCloser(
		io.NopCloser(r),
		op,
		b.opsDuration,
		b.opsFailures,
		b.isOpFailureExpected,
		nil,
		b.opsTransferredBytes,
	)
	defer trc.Close()
	err := b.bkt.Upload(ctx, name, trc)
	if err != nil {
		if !b.isOpFailureExpected(err) && ctx.Err() != context.Canceled {
			b.opsFailures.WithLabelValues(op).Inc()
		}

		return err
	}
	b.lastSuccessfulUploadTime.SetToCurrentTime()
	b.opsDuration.WithLabelValues(op).Observe(time.Since(start).Seconds())

```

objstore.go
```diff
@@ -593,14 +594,29 @@ func (b *metricBucket) Upload(ctx context.Context, name string, r io.Reader) err
...
+	if err != nil {
 		if !b.isOpFailureExpected(err) && ctx.Err() != context.Canceled {
 			b.opsFailures.WithLabelValues(op).Inc()
 		}
+
+		trc.Close()
 		return err
 	}
 	b.lastSuccessfulUploadTime.SetToCurrentTime()
 	b.opsDuration.WithLabelValues(op).Observe(time.Since(start).Seconds())
```
simonswine, line 616: As newTimingReadCloser is already recording the duration we can remove this line (not covered by any tests)

```suggestion

```

CHANGELOG.md
```diff
@@ -27,6 +27,8 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#61](https://github.com/thanos-io/objstore/pull/61) Add OpenTelemetry TracingBucket.
     > This also changes the behaviour of `client.NewBucket`. Now it returns, uninstrumented and untraced bucket.
     You can combine `objstore.WrapWithMetrics` and `tracing/{opentelemetry,opentracing}.WrapWithTraces` to have old behavior.
+- [#69](https://github.com/thanos-io/objstore/pull/69) Add `objstore_bucket_operation_transferred_bytes` that counts the number of total bytes read from the bucket operation Get/GetRange.
```
simonswine, line 30: Let's just keep the one line


```suggestion
```

simonswine (approved):
@ritaCanavarro thanks for looking into this, I another pass and I noticed a few little things, but I think it is time to get more eyes on this

objstore.go
ritaCanavarro: Done.

CHANGELOG.md
ritaCanavarro: Done.

objstore.go
ritaCanavarro: Done.

objstore.go
```diff
@@ -458,11 +458,12 @@ func WrapWithMetrics(b Bucket, reg prometheus.Registerer, name string) *metricBu
 		bkt.opsDuration.WithLabelValues(op)
 		bkt.opsFetchedBytes.WithLabelValues(op)
 	}
-	// fetched bytes only relevant for get and getrange
+
+	// fetched bytes only relevant for get, getrange and upload
```
fpetkovski, line 462:
```suggestion
	// Fetched bytes are only relevant for Get, GetRange and Upload.
```

objstore.go
```diff
@@ -692,7 +705,10 @@ func (rc *timingReadCloser) Close() error {
 
 func (rc *timingReadCloser) Read(b []byte) (n int, err error) {
 	n, err = rc.ReadCloser.Read(b)
-	rc.fetchedBytes.WithLabelValues(rc.op).Add(float64(n))
+	if rc.fetchedBytes != nil {
```
fpetkovski, line 708: When would `fetchedBytes` be nil?

objstore.go
```diff
@@ -592,15 +593,27 @@ func (b *metricBucket) Upload(ctx context.Context, name string, r io.Reader) err
...
+		b.opsTransferredBytes,
+	)
+	defer trc.Close()
+	err := b.bkt.Upload(ctx, name, trc)
+	if err != nil {
 		if !b.isOpFailureExpected(err) && ctx.Err() != context.Canceled {
 			b.opsFailures.WithLabelValues(op).Inc()
 		}
+
+		trc.Close()
```
fpetkovski, line 612: Is the `Close` call here necessary given that we have a `defer Close` above?

ritaCanavarro: Not necessary, will remove.

objstore.go
ritaCanavarro: Applied.

objstore.go
ritaCanavarro: When we use the newTimingReadCloser structure in the Upload method:
trc := newTimingReadCloser(
		io.NopCloser(r),
		op,
		b.opsDuration,
		b.opsFailures,
		b.isOpFailureExpected,
		nil,
		b.opsTransferredBytes,
	)

objstore.go
```diff
@@ -592,15 +593,25 @@ func (b *metricBucket) Upload(ctx context.Context, name string, r io.Reader) err
...
+	defer trc.Close()
+	err := b.bkt.Upload(ctx, name, trc)
+	if err != nil {
 		if !b.isOpFailureExpected(err) && ctx.Err() != context.Canceled {
 			b.opsFailures.WithLabelValues(op).Inc()
 		}
 		return err
 	}
 	b.lastSuccessfulUploadTime.SetToCurrentTime()
-	b.opsDuration.WithLabelValues(op).Observe(time.Since(start).Seconds())
```
fpetkovski, line 603: I assume we remove this because the timingReadCloser will already emit the metric.

fpetkovski (approved):
Looks good from my side, thank you 🎖️

objstore.go
ritaCanavarro: Exactly. https://github.com/thanos-io/objstore/pull/66#discussion_r1287423111

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#61](https://github.com/thanos-io/objstore/pull/61) Add OpenTelemetry TracingBucket.
     > This also changes the behaviour of `client.NewBucket`. Now it returns, uninstrumented and untraced bucket.
     You can combine `objstore.WrapWithMetrics` and `tracing/{opentelemetry,opentracing}.WrapWithTraces` to have old behavior.
+- [#69](https://github.com/thanos-io/objstore/pull/69) [#66](https://github.com/thanos-io/objstore/pull/66) Add `objstore_bucket_operation_transferred_bytes` that counts the number of total bytes read from the bucket operation Get/GetRange and also counts the number of total bytes per written to the bucket operation Upload.
```
fpetkovski, line 30:
```suggestion
- [#69](https://github.com/thanos-io/objstore/pull/69) [#66](https://github.com/thanos-io/objstore/pull/66) Add `objstore_bucket_operation_transferred_bytes` that counts the number of total bytes read from the bucket operation Get/GetRange and also counts the number of total bytes written to the bucket operation Upload.
```

objstore.go
```diff
@@ -592,15 +593,25 @@ func (b *metricBucket) Upload(ctx context.Context, name string, r io.Reader) err
 	const op = OpUpload
 	b.ops.WithLabelValues(op).Inc()
 
-	start := time.Now()
-	if err := b.bkt.Upload(ctx, name, r); err != nil {
+	trc := newTimingReadCloser(
+		io.NopCloser(r),
```
fpetkovski, line 597: We don't use this pattern for `Get` and `GetRange`. What's the reason for treating `Upload` in a special way?

ritaCanavarro: I checked and we do use the TimingReadCloser in both `Get` and `GetRange`, (Lines 543 and 565)
Also threads above where this implementation was discussed: https://github.com/thanos-io/objstore/pull/66#discussion_r1280332799 and https://github.com/thanos-io/objstore/pull/66#discussion_r1280328066

fpetkovski: What I meant was pass in a `NopCloser` into the `TimingReadCloser`, and then we close the `TimingReadCloser` ourselves, but we don't do this for `Get` and `GetRange`.

fpetkovski: Actually nevermind, we don't return a reader from the `Upload` method, so I think this makes sense.

simonswine (approved):
Yes I think this LGTM 🚀

PR #67, opened 2023-07-27
Fix overlapping metric registrations

simonswine:
This fixes a problem when multiple metrics bucket are registered with different names. Also adds a test so this won't be happening in the future.

This does not have an impact to the resulting metrics.


matej-g (approved):
Looks good to me, thanks for the fix 🙇

kakkoyun (approved):

PR #68, opened 2023-07-27
Only register fetched_bytes metrics for Get/GetRange

simonswine:
The other operations are not exposing the fetched bytes.


objstore.go
```diff
@@ -455,6 +455,11 @@ func WrapWithMetrics(b Bucket, reg prometheus.Registerer, name string) *metricBu
 		bkt.ops.WithLabelValues(op)
 		bkt.opsFailures.WithLabelValues(op)
 		bkt.opsDuration.WithLabelValues(op)
+	}
+	for _, op := range []string{
+		OpGet,
+		OpGetRange,
+	} {
```
matej-g, line 462: Should we simply put this into the other loop below with transferred bytes or do we need the extra loop?

matej-g (commented):
Looks good, just one suggestion

PR #69, opened 2023-07-27
Use a histogram to record transferred bytes for get get_range

simonswine:
This will allow to understand the distribution of object store fetches.

The metric has been purposely named so it can in the future also cover uploads. #66


bwplotka (approved):
Looks good 👍🏽

PR #70, opened 2023-08-03
GCS: Bump dependency version

matej-g:

## Changes

Updates relevant GCS dependency version and adjusts a test case to expect a new error.

This also relates to gRPC version update, see https://github.com/thanos-io/objstore/pull/60 (this change could not be merged since it requires updating the test).

## Verification

Test succeed locally.


saswatamcode (approved):
LGTM!

PR #71, opened 2023-08-16
Replace method `IsCustomerManagedKeyError` for a more generic `IsAccessDeniedErr` on the bucket interface.

alanprot:
This PR replace the `IsCustomerManagedKeyError` method for a more generic `IsAccessDeniedErr` method and implement it multiples provides.

This method can be useful for the projects consuming this lib to identify AccessDenied errors from other types of errors and for, for example, avoid retries.

I think is safe to remove the `IsCustomerManagedKeyError` for now as is a [very recent change](https://github.com/thanos-io/objstore/pull/59) and is being used only in cortex for now.

## Changes

* Remove the `IsCustomerManagedKeyError` method from the objstore interface
* Add `IsAccessDeniedErr` method on the objstore interface and implement it on multiples providers.


yeya24 (approved):
Thanks. I think this change makes sense overall. 
The only concern I have is that for different object store implementation, whether the current way of checking access deny is correct or not.
But since the method is only used in Cortex, not in Thanos at all, maybe it is fine. We can iterate on this and fix accordingly

PR #76, opened 2023-09-12
GCS: use Query.SetAttrSelection when listing objects

andyasp:

## Changes

Uses [Query.SetAttrSelection](https://pkg.go.dev/cloud.google.com/go/storage#Query.SetAttrSelection) to only populate the `Name` field in `Iter` for GCS since it's the only object field used. It is suggested in the documentation near the bottom of this [section](https://pkg.go.dev/cloud.google.com/go/storage?utm_source=godoc#hdr-Listing_objects). In case it confuses anyone else besides me `Prefix` isn't an option available to set ([reference](https://github.com/googleapis/google-cloud-go/blob/134e37b184cb8fcec1c93a4cd53df3cf60280dc9/storage/storage.go#L1561)) and isn't affected. 

## Verification

Locally performed recursive and non-recursive listings both with this change and without against a development bucket. The listing results were identical. I did not attempt to measure a performance impact myself.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#64](https://github.com/thanos-io/objstore/pull/64) OCI: OKE Workload Identity support.
 - [#73](https://github.com/thanos-io/objstore/pull/73) Аdded file path to erros from DownloadFile
 - [#51](https://github.com/thanos-io/objstore/pull/51) Azure: Support using connection string authentication.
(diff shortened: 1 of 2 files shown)


fpetkovski:
Oh, I wonder if this can help with https://github.com/thanos-io/thanos/issues/6707.

fpetkovski:
Let me test this out on a large bucket and we should be good to merge.

fpetkovski (approved):

MichaHoffmann:
:tada: thanks!

PR #79, opened 2023-09-26
Metrics: Wire up `objstore_bucket_operation_duration_seconds` in `iter` operations

gotjosh:
Trying to get some statistics in Mimir about the time we're spending in our `iter` operations I realised that despite these operations showing up as part of the labels their value was always 0.

Soon enough and with the help of Peter, I realised that these were not wired up in `objstore.go`.

This PR does exactly that and measures the duration of the whole `iter` operation. There's one big caveat though, `iter` operations take a callback as an argument which is executed per each item returned. The computed duration will include whatever time we spend on each of these callbacks which might an undesirable effect for users of the library.

Sadly, I can't get around this without changing the interface for `iter` to (e.g.) return JUST the time the network operation took. In the meantime, I've done the best I've could without incurring in a big change and documented the behaviour as part of the Histogram's HELP.

## Verification

I've used a test which I then deleted because the histogram is based on time and we don't seem to mock time so it's not worth committing.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -27,6 +27,7 @@ require (
 	golang.org/x/oauth2 v0.4.0
 	golang.org/x/sync v0.1.0
 	google.golang.org/api v0.103.0
+	google.golang.org/grpc v1.53.0
 	gopkg.in/alecthomas/kingpin.v2 v2.2.6
 	gopkg.in/yaml.v2 v2.4.0
 )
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -27,6 +27,7 @@ require (
 	golang.org/x/oauth2 v0.4.0
 	golang.org/x/sync v0.1.0
 	google.golang.org/api v0.103.0
+	google.golang.org/grpc v1.53.0
```
gotjosh, line 30: Unsure if this was forgotten or specific to my setup - happy to remove it if undesired.

MichaHoffmann (approved):
lgtm

yeya24 (approved):
Changes LGTM. I want to verify the go mod change myself first before merging it

go.mod
yeya24: Looks like you are correct. This might be a miss on previous prs

PR #85, opened 2023-11-08
s3: Allow turning off checksums in S3 Puts

sinkingpoint:
https://github.com/thanos-io/thanos/pull/6746 bumped objstore, which bumped minio, which made Thanos incompatible with otherwise compliant S3 backends that do not support the x-amz-checksum header.

## Changes

- Add in a `PutWithMD5` (open to naming changes) config entry to the S3 bucket config that allows turning off this header so that Thanos supports these backends.

## Verification

Running this at Cloudflare on our [r2](https://developers.cloudflare.com/r2/) backed Thanos that doesn't support this header

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#73](https://github.com/thanos-io/objstore/pull/73) Аdded file path to erros from DownloadFile
 - [#51](https://github.com/thanos-io/objstore/pull/51) Azure: Support using connection string authentication.
 - [#76](https://github.com/thanos-io/objstore/pull/76) GCS: Query for object names only in `Iter` to possibly improve performance when listing objects.
+- [#85](https://github.com/thanos-io/objstore/pull/85) S3: Allow checksum algorithm to be configured
 
 ### Changed
 - [#38](https://github.com/thanos-io/objstore/pull/38) *: Upgrade minio-go version to `v7.0.45`.
--- a/providers/s3/s3.go
+++ b/providers/s3/s3.go
@@ -113,6 +113,7 @@ var DefaultConfig = Config{
 	},
 	PartSize:         1024 * 1024 * 64, // 64MB.
 	BucketLookupType: AutoLookup,
+	SendContentMd5:   true, // Default to using MD5.
 }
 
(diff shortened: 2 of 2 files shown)


sinkingpoint:
@yeya24 @andyasp Mind giving this a look when you get a sec?

yeya24 (approved):
Generally it looks good. I am okay with the naming but let's see if other maintainers have other suggestions.

It would be better if we can have some doc/comment saying what is this configuration for and what S3 compatible backends might need to turn on/off this flag.

douglascamata:
I am slightly worried that there could be many people out there facing this problem and they won't easily notice. 

I think we should include the md5 always and by default to try to avoid these situations. I personally don't foresee any issues with this.

Do you think it could cause problems, @sinkingpoint @yeya24?

mlech-reef:
This problem also exists for Backblaze B2. This change would be super-useful as currently we can not use Thanos with B2 and multipart.

Forcing MD5 is also ok for us, but I guess someone that use AWS S3 could benefit from `x-amz-checksum-*` feature.

mlech-reef:
IMHO `SendContentMD5` as it is called in minio is more consistent with AWS S3 documentation - https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html#checking-object-integrity-md5

sinkingpoint:
Updated the name to be the same as minio. That makes sense - its a lot more explicit

sinkingpoint:
Defaulting to true, to disable the header by default seems like a good idea - technically speaking that would restore the previous behavior because otherwise the prior PR was a breaking change

yeya24:
I see it is still set to false. Do we want to make it default to true before I merge this

douglascamata:
I'm not a big fan of how they handle these things over at Minio: https://github.com/minio/minio-go/issues/1791. I'm 100% in favor of this project keeping the compatibility with as many s3-compatible object storage providers as possible (that's the whole purpose) without requiring users to manually disable some "random" flag because it messes up everything else besides s3 and minio itself..

sinkingpoint:
Added the `true` default, and a CHANGELOG entry

douglascamata (approved):

yeya24 (approved):

adamlounds:
for reference, Cloudflare r2 doesn't seem to need this any more

PR #86, opened 2023-11-09
GCS: Adds HTTP Config similar to S3

JoaoBraveCoding:

## Changes

Adds HTTP Config to GCS. This would be required to allow us to support hedging while using this project as a storage backend. HTTP Config allows us to configure a Transport which would enable the utilization of https://github.com/cristalhq/hedgedhttp when using the GCS provider

Related with: https://github.com/grafana/loki/pull/11132


saswatamcode (approved):
Thanks, this looks good to me overall!
Maybe @fpetkovski can take a look as well, as he's familiar with GCS.

exthttp/transport.go
```diff
@@ -11,6 +11,16 @@ import (
 	"github.com/prometheus/common/model"
 )
 
+var DefaultHTTPConfig = HTTPConfig{
```
fpetkovski, line 14: I think this might lead to an unexpected breaking change with every other provider except S3. Do we know what the defaults for GCS currently are?

JoaoBraveCoding: IIUC nowadays how objstorage works then it will eventually call `defaultBaseTransport` on the google-api-go-client in [transport/http/dial.go](https://github.com/googleapis/google-api-go-client/blob/main/transport/http/dial.go#L155-L159) which actually uses `http.DefaultTransport` [src/net/http/transport.go](https://github.com/golang/go/blob/master/src/net/http/transport.go#L38-L54), they only overwrite `MaxIdleConnsPerHost` to 100 (same value we use).

So looking at the differences if we were to move forward with my change we would be setting: 

- `MaxConnsPerHost` Since this is not set on the DefaultTransport struct I believe it will just use default `int` value of `0` which would match our value
- `ResponseHeaderTimeout` Again IIUC here the default here would be `0` which would differ from our "default" of 2 minutes.
- `DialContext.DualStack`  This apparently has been [deprecated](https://github.com/golang/go/blob/master/src/net/dial.go#L98-L105) so we can even remove it from our "default" since it's enabled by default 

And we would not be setting `ForceAttemptHTTP2` which http.DefaultTransport sets. This is where we could break some things.

With all this being said I don't have much experience with all this so I'm happy to play it safe and change the PR to make GCS use `http.DefaultTransport`

fpetkovski: Thanks for the detailed explanation 👍 Let's try this out and see if it has any significant impact.

fpetkovski (approved):

saswatamcode (approved):
Let's merge and try it out.

providers/gcs/gcs.go
```diff
@@ -76,12 +94,38 @@ func NewBucketWithConfig(ctx context.Context, logger log.Logger, gc Config, comp
 			return nil, errors.Wrap(err, "failed to create credentials from JSON")
 		}
 		opts = append(opts, option.WithCredentials(credentials))
+	} else {
+		opts = append(opts, option.WithoutAuthentication())
```
bboreham, line 98: This change is described as "Fix tests"; it stops the GCS client from using config from environment variables.

k24dizzle: +1, From what I understand, this would break GCP workload identity from working

Context: https://cloud-native.slack.com/archives/CK5RSSC10/p1633961188039700

TaylorMutch: Literally came to this repo to report this bug; this just hit me in a pre-prod environment and came to report it here :)

fpetkovski: @JoaoBraveCoding would you have time to look into this?

JoaoBraveCoding: Had plans to fix it this morning but @saswatamcode beat me to it (thank you!) Apologies for having broken GCS authentication with env vars!

PR #92, opened 2023-12-06
Enable using gRPC client for GCS

fpetkovski:
GCP has an experimental feature which allows using gRPC when talking to GCS. For the time being, it requires buckets to be explicitly added to an allowlist, but eventually the feature will become GA.

This commit allows using a gRPC client when creating a GCS bucket by toggling a boolean flag in the bucket config.

## Changes

* Allow using gRPC for GCS buckets.

### DIFF
--- a/providers/gcs/gcs.go
+++ b/providers/gcs/gcs.go
@@ -33,6 +33,7 @@ const DirDelim = "/"
 type Config struct {
 	Bucket         string `yaml:"bucket"`
 	ServiceAccount string `yaml:"service_account"`
+	UseGRPC        bool   `yaml:"use_grpc"`
 }
 
 // Bucket implements the store.Bucket and shipper.Bucket interfaces against GCS.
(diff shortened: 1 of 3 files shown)


MichaHoffmann (approved):
lgtm

providers/gcs/gcs.go
```diff
@@ -31,8 +32,10 @@ const DirDelim = "/"
 
 // Config stores the configuration for gcs bucket.
 type Config struct {
-	Bucket         string `yaml:"bucket"`
-	ServiceAccount string `yaml:"service_account"`
+	Bucket           string `yaml:"bucket"`
+	ServiceAccount   string `yaml:"service_account"`
+	UseGRPC          bool   `yaml:"use_grpc"`
+	GRPCConnPoolSize int    `yaml:"grpc_conn_pool_size"`
```
fpetkovski, line 38: I did a bit of testing by tweaking this parameter up and down. Performance follows a U-shaped curve where using 1 connection is abysmally slow, even slower than HTTP. With a pool size of 512 performance is better, but it seems to be best with 256. 

We could consider using that (or some other number) as the default to make sure users do not need to guess in the beginning. We can still keep the value configurable if anyone needs further customization.

fpetkovski: Turns out we had to enable direct path as documented here https://cloud.google.com/go/docs/reference/cloud.google.com/go/storage/latest#hdr-Experimental_gRPC_API.

```
If the application is running within GCP, users may get better performance by enabling DirectPath (enabling requests to skip some proxy steps). To enable, set the environment variable GOOGLE_CLOUD_ENABLE_DIRECT_PATH_XDS=true and add the following side-effect imports to your application:
```

```go
import (
    _ "google.golang.org/grpc/balancer/rls"
    _ "google.golang.org/grpc/xds/googledirectpath"
)
```

This made the connection pool config obsolete (as a matter of fact, anything higher than 1 would hang the client).

fpetkovski:
@saswatamcode any thoughts on this one?

providers/gcs/gcs.go
saswatamcode: We can probably mention the direct path option as a comment?

saswatamcode (approved):
Thanks! I think this generally looks good!

providers/gcs/gcs.go
fpetkovski: Good point, updated.

PR #94, opened 2023-12-29
Allow timingReadCloser to be seeker

danielblando:

## Changes
Cortex uses objstore to update blocks to s3. Cortex had added a retry functionality using seeker objects.
https://github.com/cortexproject/cortex/blob/master/pkg/storage/bucket/s3/bucket_client.go#L176-L191
It checks if an object is seeker and if so it allow it to retry. 

It also wrap the bucket store with metrics
https://github.com/cortexproject/cortex/blob/master/pkg/storage/bucket/client.go#L145

With the new timing metrics, ojbstore now override the object and thus change it to always be io.ReadCloser which is what the implementation requires, but it does not allow cortex to retry request anymore.
The change allow it to continue using a seeker object if the original object was seekable.

## Verification
Added unit tests to verify the functionality still works.


brancz:
lgtm, thank you!

PR #95, opened 2024-01-05
azure: fix when endpoint is empty

kartikaysaxena:

## Changes

Check if endpoint provided is empty in while validating the configuration


kartikaysaxena:
PTAL @yeya24 @bwplotka

squat (approved):
LGTM

GiedriusS (commented):
I'm not sure if this is correct? Shouldn't endpoint be set to the default value if it is `""`?

kartikaysaxena:
> I'm not sure if this is correct? Shouldn't endpoint be set to the default value if it is `""`?

Thanks for the suggestion! Taken into account.

PR #98, opened 2024-01-26
Fix s3 upload performance regression

pracucci:
## The symptom

The PR https://github.com/thanos-io/objstore/pull/66 introduced a regression in the S3 upload performance. After that commit, S3 uploaded are significantly slower. Here you can see how the upload latency changed for one of our Grafana Mimir clusters running in AWS:

Screenshot 2024-01-26 at 14 46 33

_Change deployed to prod on 10/06, where the latency increase. Since the upload performance in Mimir is not critical (e.g. doesn't affect neither read and write path latency), it was undetected for quite some time. Until today._

## The cause

Why the regression? Minio has some optimizations when the reader implements `io.ReaderAt` because it can leverage on `.ReadAt()`. We're used to call `Upload()` with `os.File` as a reader (I think Thanos does the same). After https://github.com/thanos-io/objstore/pull/66 wrapped the reader with the timing reader, which doesn't implement `io.ReaderAt` and so Minio couldn't optimize the upload anymore.

## The fix

In this PR I propose a fix to the issue, introducing `timingReaderSeekerReaderAt` which embeds `timingReaderSeeker`, which in turn embeds `timingReader`.

To simplify the code, I've removed the usage of `nopSeekerCloserWithSize()` and `NopCloserWithSize()` in the `Upload` and pushed down to `timingReader` whether we want it to `Close()` the wrapped reader or not. The renaming from `timingReadCloser` to `timingReader` follows this change, given it takes in input `io.Reader`.

## Manual test

I've manually that this PR fixes the upload speed regression, running a test in AWS which uploads 1GB object to S3 multiple times.

#### Before this PR

S3 bucket **not** wrapped with metrics:

```
With metrics: false Elapsed: 2.262466521s
With metrics: false Elapsed: 2.328912616s
With metrics: false Elapsed: 2.312762143s
```

S3 bucket wrapped with metrics (see higher latency):

```
With metrics: true Elapsed: 9.590695888s
With metrics: true Elapsed: 8.402357109s
With metrics: true Elapsed: 7.858041235s
```

#### After this PR

S3 bucket **not** wrapped with metrics:

```
With metrics: false Elapsed: 2.767424849s
With metrics: false Elapsed: 2.066829353s
With metrics: false Elapsed: 2.114329674s
```

S3 bucket wrapped with metrics (see **no** higher latency):

```
With metrics: true Elapsed: 2.728111635s
With metrics: true Elapsed: 2.216537896s
With metrics: true Elapsed: 2.392596594s
```

### DIFF
--- a/objstore.go
+++ b/objstore.go
@@ -232,19 +232,6 @@ func NopCloserWithSize(r io.Reader) io.ReadCloser {
 	return nopCloserWithObjectSize{r}
 }
 
-type nopSeekerCloserWithObjectSize struct{ io.Reader }
-
-func (nopSeekerCloserWithObjectSize) Close() error                 { return nil }
-func (n nopSeekerCloserWithObjectSize) ObjectSize() (int64, error) { return TryToGetSize(n.Reader) }
-
-func (n nopSeekerCloserWithObjectSize) Seek(offset int64, whence int) (int64, error) {
-	return n.Reader.(io.Seeker).Seek(offset, whence)
-}
-
-func nopSeekerCloserWithSize(r io.Reader) io.ReadSeekCloser {
-	return nopSeekerCloserWithObjectSize{r}
-}
-
 // UploadDir uploads all files in srcdir to the bucket with into a top-level directory
 // named dstdir. It is a caller responsibility to clean partial upload in case of failure.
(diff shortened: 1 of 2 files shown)


danielblando (approved):

MichaHoffmann (approved):
lgtm

yeya24 (approved):

PR #99, opened 2024-02-06
Swift: Add HTTP_Config

btaani:

## Changes

Adding HTTP Config to swift, similar to other providers, in order to support hedging & TLS
cc @JoaoBraveCoding


JoaoBraveCoding:
@btaani it seems you did not sign your commits, that's why the DCO tool is complaining

JoaoBraveCoding:
For maintainers: @btaani PR is contextualized as part of the effort we are doing to try and move Loki objstore clients to use thanos.io/objstore

saswatamcode (approved):
Thanks! LGTM! This follows #86 I think!

PR #105, opened 2024-02-28
GCS: Fix logic for allowing Google default logic in gcs provider

### LINKED ISSUE #104: Enable GCS to work with GKE workload identity
https://github.com/thanos-io/objstore/pull/86#discussion_r1505128594 I believe that this line would prevent the storage client from being able to use the environment variable `GOOGLE_APPLICATION_CREDENTIALS`. If the `service_account` field is not provided, there is no way to authenticate anymore via GKE workload identity (which was previously supported).

More context: https://cloud-native.slack.com/archives/CK5RSSC10/p1633961188039700
https://cloud-native.slack.com/archives/CK5RSSC10/p1708022085482559
https://pkg.go.dev/cloud.google.com/go/storage#NewClient

TaylorMutch:
This PR adds back logic that allows GKE workload identity to work. Fixes #104 

## Changes

* Fix logic for allowing Google default logic in gcs provider (GKE workload identity was broken due to setting anonymous auth in GCS provider if not explicitly provider service account credential https://github.com/thanos-io/objstore/pull/86/files#diff-6f1f01a82332f5074843c222ea2e28118de43aa1a516848fe5b7ef3ac18f8470R97-R98)

## Verification

I leverage this library in one of my applications to operate on objects in GCS, and when using my branch this works on my local machine when authing to GCS.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 ## Unreleased
 
 ### Fixed
+- [#105](https://github.com/thanos-io/objstore/pull/105) GCS: Fix logic for allowing Google default logic in gcs provider
 - [#102](https://github.com/thanos-io/objstore/pull/102) Azure: bump azblob sdk to get concurrency fixes.
 - [#33](https://github.com/thanos-io/objstore/pull/33) Tracing: Add `ContextWithTracer()` to inject the tracer into the context.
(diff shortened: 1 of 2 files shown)


saswatamcode:
Sorry didn't realize you already raised this, but I opened one with passing tests here: https://github.com/thanos-io/objstore/pull/106.

TaylorMutch:
Oh thanks I mentioned in the issue I wasn't sure why the original test was failing. All good, happy to see this fixed!

PR #106, opened 2024-02-28
GCS: Fix logic for falling back to Google defaults

### LINKED ISSUE #104: Enable GCS to work with GKE workload identity

saswatamcode:

## Changes

Fixes #104

### DIFF
--- a/providers/gcs/gcs.go
+++ b/providers/gcs/gcs.go
@@ -94,8 +94,6 @@ func NewBucketWithConfig(ctx context.Context, logger log.Logger, gc Config, comp
 			return nil, errors.Wrap(err, "failed to create credentials from JSON")
 		}
 		opts = append(opts, option.WithCredentials(credentials))
-	} else {
-		opts = append(opts, option.WithoutAuthentication())
 	}
 
 	opts = append(opts,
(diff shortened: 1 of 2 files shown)


JoaoBraveCoding (approved):
lgtm

jhalterman:
Unfortunately this didn't resolve the issue for us. When dropping the latest Thanos with this PR into our app, with no other changes, we're getting:
```
oauth2: cannot fetch token: 400 Bad Request\nResponse: {\"error\":\"invalid_scope\",\"error_description\":\"Invalid OAuth scope or ID token audience provided.\"
```

saswatamcode:
Could you provide more details as to how it was failing before? Was it the same error?

jhalterman:
This is a new error that only happens after we updated our objstore dependency to include the changes from #86 and this PR.

PR #109, opened 2024-03-04
GCS: adds scope to fix bug introduced in HTTPConfig support

JoaoBraveCoding:

## Changes

Adds default scope to fix the issue reported in https://github.com/thanos-io/objstore/pull/106#issuecomment-1971490145


JoaoBraveCoding:
@jhalterman This PR should fix the issue you reported, if you have the time to validate it it would be great!

jhalterman:
This seems to resolve the previous issue. Thanks @JoaoBraveCoding.

saswatamcode (approved):
Thanks a lot!

PR #112, opened 2024-03-11
s3: add DisableDualstack to config

narqo:

## Changes

Relates to https://github.com/grafana/mimir/issues/7382.

Due to internal policies, enterprises deploy VPC Interface Endpoints with Private DNS option, which overrides `s3.region.amazonaws.com`, pointing it to the local endpoint. Because S3 VPC Endpoints do not support dualstack, this adds extra requirements to the infrastructure (refer to https://github.com/minio/minio-go/issues/1766#issuecomment-1943297606 for details).

This PR adds a new configuration to `s3` client, allowing to disable the client from using AWS S3 dual-stack endpoints.


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/thanos-io/objstore
 
-go 1.20
+go 1.21
```
narqo, line 3: Note, this one comes from latest `minio-go/v7`, which is not ideal, but shouldn't be a big deal

```
go: github.com/minio/minio-go/v7@v7.0.69 requires go@1.21, but 1.20 is requested
```

matej-g (commented):
Thanks for this @narqo! Looks good to me, but it seems like we'll also need to bump the Go version in our setup. Would you mind bumping the version in the `.go-version` file to ensure also our CI passes?

narqo:
@matej-g could you have a glance over the latest changes. Is there anything I can do to help with landing them to mainline?

matej-g (approved):
Sorry for the long delay @narqo, this looks good to me, thanks you 🙇

PR #115, opened 2024-04-03
GCS: Fix creation of bucket with GRPC connections

pedro-stanaka:
## Summary

Some recent changes introduced a regression for folks using GRPC connections to GCS.

In this PR I am fixing that issue and introducing test covering creation of the bucket using the **public** constructor.


go.mod
```diff
@@ -3,14 +3,15 @@ module github.com/thanos-io/objstore
 go 1.20
 
 require (
-	cloud.google.com/go/storage v1.35.1
+	cloud.google.com/go/storage v1.40.0
```
fpetkovski, line 6: Do we need to update the library? Maybe we can add this to CHANGELOG.

providers/gcs/gcs.go
```diff
@@ -119,16 +131,14 @@ func NewBucketWithConfig(ctx context.Context, logger log.Logger, gc Config, comp
 	opts = append(opts, option.WithScopes(storage.ScopeFullControl, "https://www.googleapis.com/auth/cloud-platform"))
 	gRT, err := htransport.NewTransport(context.Background(), rt, opts...)
 	if err != nil {
-		return nil, err
+		return nil, nil
```
fpetkovski, line 134: Typo?

pedro-stanaka: yep, refactor typo

go.mod
pedro-stanaka: This got pulled in because of peer libraries (gcsemu). But downstream users still can pin their versions.

providers/gcs/gcs_test.go
```diff
@@ -13,8 +13,10 @@ import (
 	"time"
 
 	"github.com/efficientgo/core/testutil"
+	"github.com/fullstorydev/emulators/storage/gcsemu"
 	"github.com/go-kit/log"
 	"github.com/prometheus/common/model"
+	"github.com/stretchr/testify/require"
```
fpetkovski, line 19: There is already `testutil`.

pedro-stanaka: Ah, this was copy-pasta error. Thx

fpetkovski (approved):
Nice, thanks for the fix.

PR #116, opened 2024-04-17
azure: Add new storage_create_container configuration property

willyborankin:
In some cases account does not have permissions to read container properties or create one but has all permissions to do CRUD operations inside the container, e.g SAS tokens. To solve such use case the new configuration property was added for the Azure object storage config which creates a new container explicitly by settings `storage_create_container` to `true` or `false`. To keep backward compatibility with existing object storage configurations the default value for it is always `true`.

## Changes
azure: Add new storage_create_container configuration property


MichaHoffmann (approved):
LGTM

MichaHoffmann (approved):
lgtm

saswatamcode (approved):

PR #117, opened 2024-05-01
metrics: don't increment `objstore_bucket_operation_failures_total` if context cancelled while reading

charleskorn:

## Changes

This PR fixes an issue where the `objstore_bucket_operation_failures_total` metric is incremented if the context is cancelled while reading an object from a bucket but after the `Get` or `GetRange` method has returned.

This makes the behaviour consistent with other operations, and makes the behaviour consistent with a cancellation that occurs while `Get` or `GetRange` is executing.

## Verification

I have added unit tests to cover this scenario.

### DIFF
--- a/objstore.go
+++ b/objstore.go
@@ -551,6 +551,7 @@ func (b *metricBucket) Get(ctx context.Context, name string) (io.ReadCloser, err
 		return nil, err
 	}
 	return newTimingReader(
+		ctx,
 		rc,
 		true,
 		op,
@@ -574,6 +575,7 @@ func (b *metricBucket) GetRange(ctx context.Context, name string, off, length in
 		return nil, err
 	}
 	return newTimingReader(
+		ctx,
 		rc,
 		true,
 		op,
@@ -606,6 +608,7 @@ func (b *metricBucket) Upload(ctx context.Context, name string, r io.Reader) err
 	b.ops.WithLabelValues(op).Inc()
 
 	trc := newTimingReader(
+		ctx,
 		r,
 		false,
 		op,
@@ -663,6 +666,8 @@ func (b *metricBucket) Name() string {
 type timingReader struct {
 	io.Reader
 
+	ctx context.Context
+
 	// closeReader holds whether the wrapper io.Reader should be closed when
 	// Close() is called on the timingReader.
 	closeReader bool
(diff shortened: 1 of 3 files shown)


MichaHoffmann (commented):
Couldn't we get the same by marking canceled context as expected error?

charleskorn:
> Couldn't we get the same by marking canceled context as expected error?

Yes. However, cancelled context is currently treated as a special case everywhere else that checks for expected errors (eg. [here](https://github.com/thanos-io/objstore/blob/a1e58aa3bed5e446d4d66c8ce27973dc24d00850/objstore.go#L548) and [here](https://github.com/thanos-io/objstore/blob/a1e58aa3bed5e446d4d66c8ce27973dc24d00850/objstore.go#L533)), so this PR makes reading an object consistent with that. 

We could remove all the special checks for context cancellation, but that would then be a change in current behaviour for all consumers of this library.

objstore.go
```diff
@@ -756,7 +762,7 @@ func (r *timingReader) Read(b []byte) (n int, err error) {
 	r.readBytes += int64(n)
 	// Report metric just once.
 	if !r.alreadyGotErr && err != nil && err != io.EOF {
-		if !r.isFailureExpected(err) {
+		if !r.isFailureExpected(err) && r.ctx.Err() != context.Canceled {
```
pracucci, line 765: Have you considered using `errors.Is()` here?

pracucci (approved):
LGTM (modulo a minor commenet)

objstore.go
charleskorn: The other places that apply a similar special case check (eg. [here](https://github.com/thanos-io/objstore/blob/a1e58aa3bed5e446d4d66c8ce27973dc24d00850/objstore.go#L548) and [here](https://github.com/thanos-io/objstore/blob/a1e58aa3bed5e446d4d66c8ce27973dc24d00850/objstore.go#L533)) don't use `errors.Is()`, so I did the same thing here for consistency.

fpetkovski: I find it strange that the only reason for injecting the context is to check if it got cancelled, but the context is not used for the read operation. In practice a user can inject any context, even one which is not associated with the read.

Isn't the reader going to return `context.Cancelled` if the `Read` fails due to cancellation?

charleskorn: Fair point - I've changed this in d44110f to check that the error returned satisfies `errors.Is(err, context.Canceled)`.

fpetkovski (approved):

MichaHoffmann (approved):
lgtm

MichaHoffmann:
Isn't this actually hiding an error? If the initial s3 read that we do to check for 404 fails because of connection issue we don't get timing or error metrics

PR #118, opened 2024-05-04
[WIP] Hedged-request: Handling in HTTP Client Configuration

Vanshikav123:

## Changes

Initiation change for
[https://github.com/thanos-io/thanos/issues/6712](https://github.com/thanos-io/thanos/issues/6712)

### DIFF
--- a/providers/gcs/gcs.go
+++ b/providers/gcs/gcs.go
@@ -15,6 +15,7 @@ import (
 	"time"
 
 	"cloud.google.com/go/storage"
+	"github.com/cristalhq/hedgedhttp"
 	"github.com/go-kit/log"
 	"github.com/pkg/errors"
 	"github.com/prometheus/common/version"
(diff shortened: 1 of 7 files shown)


Vanshikav123:
@GiedriusS PTAL!

providers/gcs/gcs.go
```diff
@@ -135,8 +136,15 @@ func appendHttpOptions(gc Config, opts []option.ClientOption) ([]option.ClientOp
 		return nil, err
 	}
 
-	httpCli := &http.Client{
+	hedgedClient, err := hedgedhttp.New(hedgedhttp.Config{
```
GiedriusS, line 139: All of this `hedgedhttp.New()` must be inside Thanos code, not here, because otherwise you will be forcing hedged HTTP requests on all clients of this library.

providers/s3/s3.go
```diff
@@ -227,12 +228,18 @@ func NewBucketWithConfig(logger log.Logger, config Config, component string) (*B
...
+			Transport: http.DefaultTransport,
+		})
+		if err != nil {
+			return nil, err
+		}
 		chain = []credentials.Provider{
 			wrapCredentialsProvider(&credentials.EnvAWS{}),
 			wrapCredentialsProvider(&credentials.FileAWSCredentials{}),
 			wrapCredentialsProvider(&credentials.IAM{
 				Client: &http.Client{
```
GiedriusS, line 241: The Transport need to come from this function here: https://github.com/thanos-io/objstore/blob/main/client/factory.go#L65

GiedriusS (changes requested):
Thanks for your work on this :bow: I added suggestions in line. Please let me know if you have any questions

providers/gcs/gcs.go
Vanshikav123: Thank you for your assistance ! i want to confirm my query that you want me to create hedgedhttp.Client into the Thanos codebase itself, and expose a configuration option only to enable or disable hedged requests? something like this ?

Vanshikav123:
Hey ! @GiedriusS closing this one ! PTAL for #118

PR #119, opened 2024-05-19
Hedged-request: Handling in HTTP Client Configuration

Vanshikav123:

## Changes

Initiation change for
[https://github.com/thanos-io/thanos/issues/6712](https://github.com/thanos-io/thanos/issues/6712)

### DIFF
--- a/client/factory.go
+++ b/client/factory.go
@@ -6,8 +6,11 @@ package client
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"strings"
+	"time"
 
+	"github.com/cristalhq/hedgedhttp"
 	"github.com/thanos-io/objstore"
 	"github.com/thanos-io/objstore/providers/azure"
 	"github.com/thanos-io/objstore/providers/bos"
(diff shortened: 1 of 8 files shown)


providers/oci/oci.go
```diff
@@ -288,7 +288,7 @@ func (b *Bucket) deleteBucket(ctx context.Context) (err error) {
 }
 
 // NewBucket returns a new Bucket using the provided oci config values.
-func NewBucket(logger log.Logger, ociConfig []byte) (*Bucket, error) {
+func NewBucket(logger log.Logger, ociConfig []byte, rt http.RoundTripper) (*Bucket, error) {
```
GiedriusS, line 291: And where is `rt http.RoundTripper` used? :( same with all other providers

GiedriusS (changes requested):

client/factory.go
```diff
@@ -64,23 +65,23 @@ func NewBucket(logger log.Logger, confContentYaml []byte, component string) (obj
...
-		bucket, err = gcs.NewBucket(context.Background(), logger, config, component)
+		bucket, err = gcs.NewBucket(context.Background(), logger, config, component, rt)
 	case string(S3):
-		bucket, err = s3.NewBucket(logger, config, component)
+		bucket, err = s3.NewBucket(logger, config, component, rt)
 	case string(AZURE):
-		bucket, err = azure.NewBucket(logger, config, component)
+		bucket, err = azure.NewBucket(logger, config, component, rt)
 	case string(SWIFT):
 		bucket, err = swift.NewContainer(logger, config)
```
GiedriusS, line 74: It's not supported by all providers? :thinking:

Vanshikav123: > It's not supported by all providers? 🤔 I checked a few like swift/bos and I don't see a reason why this is not implemented there.

i added this parameter to the providers which had http.Client , swift does not uses NewBucket  but bos/obs can have this parameter i think also i see that obs makes new ObsClient and bos uses NewClient  which makes a BOS client...wdyt?

PR #121, opened 2024-06-11
Add support for WriterTo in TimingReader

fpetkovski:
This commit adds support for the WriterTo interface in the TimingReader so that it can be used efficiently in io.Copy.


pracucci (approved):
Nice job, LGTM!

MichaHoffmann:
If `r` implements Seeker then we wont ever return a writerTo, right? Is that problematic?

fpetkovski:
Let's merge this and we can add all combinations later.

MichaHoffmann (approved):

PR #130, opened 2024-07-29
feat: Decouple creating bucket metrics from instrumenting the bucket

kavirajk:

## Changes
Currently if a single process uses multiple object client, we get metrics duplicate registration failures.

With this PR,
1. New `BucketMetrics` API return just a `Metrics` that can be created once and used for multiple clients.
2. New `Wrap` API that takes `metrics` and returns instrumented bucket just like `WrapWithMetrics` API.

This PR doesn't change any existing API `WrapWithMetrics`

## Verification

Not breaking any existing API(s). All existing tests passes


yeya24 (approved):
Thanks!

kavirajk:
@yeya24 thanks. Can you merge it as well?

yeya24:
Done! I was hoping to get an approval from another maintainer but I think the change should be straightforward

PR #136, opened 2024-09-09
s3:Generate presigned URL

### LINKED ISSUE #135: Feature request: Bucket method to export presigned upload URL
We have use case where we need presigned URL to upload tsdb blocks directly to object storage. 
In a Bucket struct [client object](https://github.com/thanos-io/objstore/blob/main/providers/s3/s3.go#L158) is private and cannot be accessed from outside hence need separate method which returns presigned URL.

patilsuraj767:
Fixes #135


MichaHoffmann:
Why do we need objstore client to do that? Since this is outside of thanos, couldnt we just use plain old minio client? Only thing we save is that we can reuse the same configuration.

patilsuraj767:
Yes, I agree we can use plain old minio client but we are anyway creating the bucket object to perform other tasks like downloading `meta.json` of tsdb, adding thanos external label and uploading it back again. So I thought instead of creating separate minio client it would be good have a method to get presignedURL in thanos objstore.

MichaHoffmann:
But the objstore client itself does not know those files or the schema of meta.json at all; all of this is done inside of thanos using an objstore.Bucket interface but the objstore.Bucket implementation knows of nothing of this. I think this would better be done with just a minio client in an external program probably.

patilsuraj767:
Yes, I meant we use objstore for downloading and uploading of files. I was just thinking to reuse the minio client config which is already generated in objstore. But no worries if you think this is outside the scope of the project, I will generate the separate minio client.

MichaHoffmann:
> Yes, I meant we use objstore for downloading and uploading of files. I was just thinking to reuse the minio client config which is already generated in objstore. But no worries if you think this is outside the scope of the project, I will generate the separate minio client.

Oh sorry, I dont have the last say here of course! Its just that for me intuitively - this feels like it should just be minio client usage. Lets wait for other opinions!

patilsuraj767:
ok Thanks @MichaHoffmann 
Another issue is this [parseConfig](https://github.com/thanos-io/objstore/blob/main/providers/s3/s3.go#L169) method is also private. We might need to rewrite it again in our codebase to parse the s3 config yaml and then create the separate minio client.

PR #137, opened 2024-09-12
fix: timing reader ignores initial get_range

MichaHoffmann:
Some objstore providers like s3 do one initial read that already talks to object storage, this read might fail due to timeout, which is invisible in the current instrumentation since only subsequent reads are instrumented. This PR fixes this.


saswatamcode (approved):

yeya24 (approved):
Thanks!

PR #138, opened 2024-09-12
s3: stat object instead of empty read

MichaHoffmann:
Minio Client issues a HEAD request against the object which for some object storage providers might be easier then a GET with Range header set to zero bytes.


saswatamcode (approved):
Thanks, might be worth trying out!

PR #142, opened 2024-09-26
Pass server-side encryption option to minio.StatObject call

### LINKED ISSUE #141: SSE-C encryption broken for Exits/Attributes call
Hi,

I configured thanos to use `SSE-C` for encryption.

In contrast to `SSE-KMS` and default `SSE` the client has to pass the key for call, which currently is missing for `Exists()` and `Attributes()` method of the S3 client.

jabbrwcky:
Adds missing encryption data passed to `minio.StatObject()` calls, causing client methods `Exists()` and `Attributes()` to fail with a 400 Bad Request error.

This may only affects encryption using `SSE-C` because here it is the callers responsibility to send the encryption key along with the request. In contrast `SSE-KMS` and `SSE-S3` are handled transparently (if the user has permissions to use the key in the casse of `SSE-KMS`)

Fixes #141 

## Changes

* Add SSE-C configuration to `Bucket.Exists()`
* Add SSE-C configuration to `Bucket.Attributes()`

## Verification

Added E2E-Test that uploads, checks for existence and downloads a file using SSE-C

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 
 ### Fixed
 - [#153](https://github.com/thanos-io/objstore/pull/153) Metrics: Fix `objstore_bucket_operation_duration_seconds_*` for `get` and `get_range` operations.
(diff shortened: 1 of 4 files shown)


toporek3112:
Any updates on this one?

JoaoBraveCoding (approved):
lgtm

jabbrwcky:
@JoaoBraveCoding Rebasing onto main has reset the automerge for the PR

PR #143, opened 2024-09-30
S3: exporting minio client

patilsuraj767:
Exposing minio client so that it can be used for accessing other methods of minio client library. One of the use case is for generating presigned URLs.


saswatamcode (approved):
LGTM! I think this is more preferable than #136

MichaHoffmann:
I dont think this is good library design, the particular S3 client to use is an implementation detail here - this library offers no utility except configuration for this usecase. Why would one not just use minio directly instead of having a dependency on objstore. I would prefer if we revert this PR

PR #145, opened 2024-10-08
Include content length in the response of Get and GetRange

ashwanthgoli:
This pr adds implements `ObjectSize()` method on the return values of `Get` and `GetRange`. This should help readers preallocate buffers based on the content size to reduce number of allocations. They can call `TryToGetSize` on the reader to get the size, existing comment already highlights that this is best effort.

~~I did not add it for swift since the [Length call could potentially result in another request to the server](https://github.com/ncw/swift/blob/master/swift.go#L1741)~~

## Verification

Updated the acceptance test to check for object size after Get and GetRange calls

### DIFF
--- a/objstore.go
+++ b/objstore.go
@@ -829,3 +829,13 @@ func (t *timingReaderWriterTo) WriteTo(w io.Writer) (n int64, err error) {
 	t.timingReader.updateMetrics(int(n), err)
 	return n, err
 }
+
+type ObjectSizerReadCloser struct {
+	io.ReadCloser
+	Size int64
+}
+
+// ObjectSize implement ObjectSizer.
+func (o ObjectSizerReadCloser) ObjectSize() (int64, error) {
+	return o.Size, nil
+}
--- a/objstore.go
+++ b/objstore.go
@@ -837,5 +837,9 @@ type ObjectSizerReadCloser struct {
 
 // ObjectSize implement ObjectSizer.
 func (o ObjectSizerReadCloser) ObjectSize() (int64, error) {
+	if o.Size == nil {
+		return 0, errors.New("unknown size")
+	}
+
 	return o.Size()
 }
--- a/objstore.go
+++ b/objstore.go
@@ -832,10 +832,10 @@ func (t *timingReaderWriterTo) WriteTo(w io.Writer) (n int64, err error) {
 
 type ObjectSizerReadCloser struct {
 	io.ReadCloser
-	Size int64
+	Size func() (int64, error)
 }
 
 // ObjectSize implement ObjectSizer.
 func (o ObjectSizerReadCloser) ObjectSize() (int64, error) {
-	return o.Size, nil
+	return o.Size()
 }
--- a/providers/gcs/gcs.go
+++ b/providers/gcs/gcs.go
@@ -226,12 +226,22 @@ func (b *Bucket) Iter(ctx context.Context, dir string, f func(string) error, opt
 
 // Get returns a reader for the given object name.
 func (b *Bucket) Get(ctx context.Context, name string) (io.ReadCloser, error) {
-	return b.bkt.Object(name).NewReader(ctx)
+	r, err := b.bkt.Object(name).NewReader(ctx)
+	if err != nil {
+		return r, err
+	}
+
+	return objstore.ObjectSizerReadCloser{ReadCloser: r, Size: r.Attrs.Size}, nil
 }
 
 // GetRange returns a new range reader for the given object name and range.
 func (b *Bucket) GetRange(ctx context.Context, name string, off, length int64) (io.ReadCloser, error) {
-	return b.bkt.Object(name).NewRangeReader(ctx, off, length)
+	r, err := b.bkt.Object(name).NewRangeReader(ctx, off, length)
+	if err != nil {
+		return r, err
+	}
+
+	return objstore.ObjectSizerReadCloser{ReadCloser: r, Size: r.Attrs.Size}, nil
 }
 
 // Attributes returns information about the specified object.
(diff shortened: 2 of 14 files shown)


providers/gcs/gcs.go
```diff
@@ -226,12 +226,22 @@ func (b *Bucket) Iter(ctx context.Context, dir string, f func(string) error, opt
...
 
 // GetRange returns a new range reader for the given object name and range.
 func (b *Bucket) GetRange(ctx context.Context, name string, off, length int64) (io.ReadCloser, error) {
-	return b.bkt.Object(name).NewRangeReader(ctx, off, length)
+	r, err := b.bkt.Object(name).NewRangeReader(ctx, off, length)
+	if err != nil {
+		return r, err
+	}
+
+	return objstore.ObjectSizerReadCloser{ReadCloser: r, Size: r.Attrs.Size}, nil
```
fpetkovski, line 244: Should the size be set to `length` here?

fpetkovski: This is where that value is set: https://github.com/googleapis/google-cloud-go/blob/main/storage/grpc_client.go#L1126.

According to the code, this is the size of the entire object, even if a range is requested: https://github.com/googleapis/google-cloud-go/blob/main/storage/grpc_client.go#L1109-L1110.

We might want to test this better to be sure we're returning the right value.

ashwanthgoli: thanks! adding it for `GetRange` was an after thought, i'll have another look

ashwanthgoli: [`Remain()`](https://github.com/googleapis/google-cloud-go/blob/main/storage/reader.go#L270) looks like the right value to use for Range. wdyt?

https://github.com/googleapis/google-cloud-go/blob/main/storage/grpc_client.go#L1151
https://github.com/googleapis/google-cloud-go/blob/main/storage/http_client.go#L1434

providers/s3/s3.go
```diff
@@ -439,10 +439,17 @@ func (b *Bucket) getRange(ctx context.Context, name string, off, length int64) (
...
 		return nil, err
 	}
-	return b.client.GetObject(ctx, b.name, name, *opts)
+
+	o, err := b.client.GetObject(ctx, b.name, name, *opts)
+	if err != nil {
+		return o, err
+	}
+
+	return objstore.ObjectSizerReadCloser{ReadCloser: o, Size: stat.Size}, nil
```
ashwanthgoli, line 452: references:

size is taken from `Content-Length` header
https://github.com/minio/minio-go/blob/master/utils.go#L264
https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadObject.html
```
HeadObject returns only the metadata for an object. If the Range is satisfiable, only the ContentLength is affected in the response. If the Range is not satisfiable, S3 returns a 416 - Requested Range Not Satisfiable error.
```

providers/gcs/gcs.go
fpetkovski: Yes, this could be the one that we need. We have e2e tests that run in the CI which could be useful for this: https://github.com/thanos-io/objstore/blob/main/testing.go#L83.

ashwanthgoli: other providers are returning content length which i am hoping reflects the requested range, I'll try to do a basic sanity check for a couple of providers tomorrow.

if this feels risky, i can limit the change to `Get` call alone

ashwanthgoli: 👍 i'll adding something there :)

providers/swift/swift.go
```diff
@@ -262,7 +262,11 @@ func (c *Container) get(name string, headers swift.Headers, checkHash bool) (io.
 	if err != nil {
 		return nil, errors.Wrap(err, "open object")
 	}
-	return file, err
+
+	return objstore.ObjectSizerReadCloser{
+		ReadCloser: file,
+		Size:       file.Length,
```
ashwanthgoli, line 268: `file.Length()` returns the value of `Content-Length` response header if it is set, else it makes a HEAD call to the store

providers/gcs/gcs.go
ashwanthgoli: @fpetkovski added the tests, it's looking good. can you take another look when you get a chance? Thanks!

objstore.go
```diff
@@ -829,3 +829,13 @@ func (t *timingReaderWriterTo) WriteTo(w io.Writer) (n int64, err error) {
 	t.timingReader.updateMetrics(int(n), err)
 	return n, err
 }
+
+type ObjectSizerReadCloser struct {
+	io.ReadCloser
+	Size func() (int64, error)
```
fpetkovski, line 835: From what I see, we always return `nil` as the error, so I wonder why we need to have it as a return argument.

fpetkovski (commented):
Looks great, I just have one comment.

objstore.go
ashwanthgoli: swift could potentially return err if the HEAD call fails

fpetkovski (approved):
Thanks, LGTM

ashwanthgoli:
@fpetkovski can you help merge this change? :)

providers/s3/s3.go
MichaHoffmann: Isnt this adding an additional request to the GetRange call?

ashwanthgoli: minio-go [saves the objectInfo](https://github.com/minio/minio-go/blob/master/api-get-object.go#L326-L335) when we get the object. [calling Stat()](https://github.com/minio/minio-go/blob/master/api-get-object.go#L428) after a successful `GetObject()` call won't trigger the `HEAD` request i think.

this is very likely not going to cause a regression since this code executes only when some asks for `Size()`.
but let me know if you notice any problems :)

PR #150, opened 2024-10-21
Added roundtripper wrapper as param

milinddethe15:

## Changes

Added a parameter in NewBucket() to pass `wrapRoundtripper func(http.RoundTripper) http.RoundTripper`
## Verification

Created tests in which a always error returning roundtripper wrapper is passed to NewBucket() and tested whether this error is caught or not.

### DIFF
--- a/errutil/rt_error.go
+++ b/errutil/rt_error.go
@@ -1,8 +1,9 @@
 package errutil
 
 import (
-	"errors"
 	"net/http"
+
+	"github.com/pkg/errors"
 )
 
 var Rt_err = errors.New("RoundTripper error")
--- a/errutil/rt_error.go
+++ b/errutil/rt_error.go
@@ -6,7 +6,11 @@ import (
 	"github.com/pkg/errors"
 )
 
-var Rt_err = errors.New("RoundTripper error")
+var rtErr = errors.New("RoundTripper error")
+
+func IsMockedError(err error) bool {
+	return errors.Is(err, rtErr)
+}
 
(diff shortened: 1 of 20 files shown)


errutil/rt_error.go
```diff
@@ -1,6 +1,12 @@
 package errutil
 
-import "net/http"
+import (
+	"net/http"
+
+	"github.com/pkg/errors"
+)
+
+var Rt_err = errors.New("RoundTripper error")
```
GiedriusS, line 9: My suggestion would be to hide this variable and create a function for it:

```suggestion
var rtErr = errors.New("RoundTripper error")

func IsMockedError(err error) bool {
  return errors.Is(err, rtErr)
}
```

This is mostly because `Rt_err` isn't a constant (and it cannot be). This will make it more maintainable IMO

errutil/rt_error.go
```diff
@@ -10,3 +16,7 @@ type ErrorRoundTripper struct {
 func (ert *ErrorRoundTripper) RoundTrip(*http.Request) (*http.Response, error) {
 	return nil, ert.Err
 }
+
+func WrapRoundtripper(rt http.RoundTripper) http.RoundTripper {
```
GiedriusS, line 20:
```suggestion
func WrapWithErrRoundtripper(rt http.RoundTripper) http.RoundTripper {
```

GiedriusS (approved):
Two suggestions but overall great work :1st_place_medal:

PR #153, opened 2024-10-28
fix objstore_bucket_operation_duration_seconds metrics

siavashs:

## Changes

* fix histogram metrics for `get` and `getrange` operations.


MichaHoffmann (approved):

bboreham:
Thanks for fixing this!
Drive-by observation: "fix" doesn't tell me much.
"Change from nanoseconds to seconds" appears to be what happened.

PR #157, opened 2024-11-20
Azure: Add new `az_tenant_id`, `client_id` and `client_secret` configs

DylanGuedes:

## Changes

* Adds two new Azure configs: `client_secret`, `client_id`, and `az_tenant_id`
* Modifies Azure authentication: if `az_tenant_id`, `client_secret` and `client_id` are set, we use their values to generate an authentication token.

## Verification

I have created a unitary test locally that interacts with a valid existing Azure blob and I managed to authenticate correctly with the new method.


providers/azure/azure.go
```diff
@@ -46,6 +46,8 @@ var DefaultConfig = Config{
 
 // Config Azure storage configuration.
 type Config struct {
+	ClientSecret            string             `yaml:"client_secret"`
```
ashwanthgoli, line 49: when authentication using `ClientSecret`, can we rely on a new `ClientID` field instead of reusing `UserAssignedID`. I think that makes the intent explicit

DylanGuedes: good point. Pushed a commit adding it.

providers/azure/helpers.go
```diff
@@ -94,3 +83,17 @@ func getContainerClient(conf Config, wrapRoundtripper func(http.RoundTripper) ht
...
 }
+
+func getTokenCredential(conf Config) (azcore.TokenCredential, error) {
+	if conf.UserAssignedID == "" {
+		return azidentity.NewDefaultAzureCredential(nil)
+	}
+
+	if conf.ClientSecret != "" && conf.TenantID != "" && conf.ClientID != "" {
+		return azidentity.NewClientSecretCredential(conf.TenantID, conf.ClientID, conf.ClientSecret, &azidentity.ClientSecretCredentialOptions{})
+	}
```
JoaoBraveCoding, line 94: Since UserAddignedID & ClientSecret cannot be used at the same time doesn't the current implementation make it so the code in line 93 is never executed because we will always exit on 89, when `UserAssignedID` is not defined?

DylanGuedes: good catch. I'll revert the order and add a test covering it :D

JoaoBraveCoding (approved):
lgtm

providers/azure/azure.go
```diff
@@ -46,6 +46,9 @@ var DefaultConfig = Config{
 
 // Config Azure storage configuration.
 type Config struct {
+	ClientID                string             `yaml:"client_id"`
+	ClientSecret            string             `yaml:"client_secret"`
+	TenantID                string             `yaml:"tenant_id"`
```
JoaoBraveCoding, line 51: small nit: WDYT of changing the config field to one of these to avoid users possibly getting confused with `tenant_id` from either (Thanos/Loki/Mimir)?
```suggestion
	TenantID                string             `yaml:"az_tenant_id"`
```
or 

```suggestion
	TenantID                string             `yaml:"csc_tenant_id"`
```

JoaoBraveCoding (commented):
Overall lgtm, just one small nit

providers/azure/azure.go
DylanGuedes: that's fair. I'm going for `az_tenant_id` as googling that term led to better results than `csc_tenant_id`

JoaoBraveCoding (approved):
lgtm

PR #161, opened 2025-01-06
feat(gcs): Support setting endpoint for Google Cloud Storage (GCS) in Thanos

Saumya40-codes:

## Changes

Part of https://github.com/thanos-io/thanos/issues/7862

- This PR adds an Endpoint field for the `gcs`. Now if the custom endpoint is specified it will overide the default one.


Saumya40-codes:
Should tests be added for this change?

PR #162, opened 2025-01-14
Upgrade to minio v7.0.83

aknuds1:

## Changes

Upgrade to minio v7.0.83.


MichaHoffmann (approved):

providers/s3/s3_aws_sdk_auth.go
```diff
@@ -26,8 +26,7 @@ func NewAWSSDKAuth(region string) *credentials.Credentials {
 	})
 }
 
-// Retrieve retrieves the keys from the environment.
-func (a *AWSSDKAuth) Retrieve() (credentials.Value, error) {
+func (a *AWSSDKAuth) RetrieveWithCredContext(cc *credentials.CredContext) (credentials.Value, error) {
```
matej-g, line 29: What is the reason for changing the method / signature here? I might be missing some context.

matej-g (commented):
Update looks good, just a small question about the code change

providers/s3/s3_aws_sdk_auth.go
aknuds1: @matej-g it's not actually changing, the minio interface adds the method `RetrieveWithCredContext` and deprecates `Retrieve`. `Retrieve` is still there, if you look on line 50 :)

matej-g: Got it, I was wondering about the addition, but the interface change explains it. Thanks 👍

matej-g (approved):

PR #163, opened 2025-01-14
Add constructor for a client from a BucketConfig object

fpetkovski:
It's handy to be able to create a bucket from an existing BucketConfig object in addition to using raw yaml to do it.


JoaoBraveCoding (approved):
From my POV, lgtm

PR #164, opened 2025-01-29
Add support for object upload options

fpetkovski:
I would like to be able to set the content type on uploaded objects since some providers cannot show a preview if the content type is missing.

To facilitate this, I've extended the UploadObject method to accept a variable list of options. The only introduction option is the content type, but it is extensible for other options.

### DIFF
--- a/objstore.go
+++ b/objstore.go
@@ -62,7 +62,7 @@ type Bucket interface {
 
 	// Upload the contents of the reader as an object into the bucket.
 	// Upload should be idempotent.
-	Upload(ctx context.Context, name string, r io.Reader) error
+	Upload(ctx context.Context, name string, r io.Reader, opts ...ObjectUploadOption) error
 
 	// Delete removes the object with the given name.
 	// If object does not exist in the moment of deletion, Delete should throw error.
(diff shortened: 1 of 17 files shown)


objstore.go
```diff
@@ -62,7 +62,7 @@ type Bucket interface {
 
 	// Upload the contents of the reader as an object into the bucket.
 	// Upload should be idempotent.
-	Upload(ctx context.Context, name string, r io.Reader) error
+	Upload(ctx context.Context, name string, r io.Reader, opts ...ObjectUploadOption) error
```
bwplotka, line 65: Why not simpler?

```suggestion
	Upload(ctx context.Context, name string, r io.Reader, opts ...UploadOption) error
```

objstore.go
```diff
@@ -196,6 +196,26 @@ func ApplyIterOptions(options ...IterOption) IterParams {
 	return out
 }
 
+type UploadObjectParams struct {
+	ContentType string
+}
+
+type ObjectUploadOption func(f *UploadObjectParams)
```
bwplotka, line 203: I think we can simplify (no need for params to be public struct):

```suggestion
type UploadOption func(f *uploadOptions)
```

bwplotka (commented):
Some nits, otherwise LGTM!

objstore.go
fpetkovski: Looks like we already have this type: https://github.com/thanos-io/objstore/blob/3f8838fdd5a2c99bc3d771ef78816201e7d0b69e/objstore.go#L233.

It is used for the `UploadDir` function.

MichaHoffmann (approved):

fpetkovski:
Failures in CI seem unrelated to this change, so I will merge the PR.

MichaHoffmann:
We could add mime type sniffing like here https://github.com/perkeep/perkeep/blob/master/internal/magic/magic.go#L199 to the "Upload" function

PR #166, opened 2025-02-19
allow custom config in NewBucket()

milinddethe15:

## Changes
Remove strict unmarshal of confContentYaml


milinddethe15:
cc @MichaHoffmann

bwplotka (commented):
This makes this function less safe e.g. you make a typo and think you configured a field, but you didn't.

Should we instead have a quick method for trimming the custom fields?

client/factory.go
```diff
@@ -38,7 +38,7 @@ type BucketConfig struct {
 func NewBucket(logger log.Logger, confContentYaml []byte, component string, wrapRoundtripper func(http.RoundTripper) http.RoundTripper) (objstore.Bucket, error) {
 	level.Info(logger).Log("msg", "loading bucket configuration")
 	bucketConf := &BucketConfig{}
-	if err := yaml.UnmarshalStrict(confContentYaml, bucketConf); err != nil {
+	if err := yaml.Unmarshal(confContentYaml, bucketConf); err != nil {
```
milinddethe15, line 41:
```suggestion
	if err := yaml.UnmarshalStrict(confContentYaml, bucketConf); err != nil {
```
This will return error, if something is off

PR #168, opened 2025-02-27
Add trimmer for extra field in config yaml

milinddethe15:

## Changes
New function to trim extra fields from YAML config. 

* [`clientutil/trimconfig.go`](diffhunk://#diff-dd1cabacc4deb1cb3beac6ea61af914d5534a5611a3c56a502c52ffcbe5661baR1-R28): Added a new function `TrimExtraFields` that removes any fields from the YAML configuration that are not explicitly allowed.

## Verification
* [`clientutil/trimconfig_test.go`](diffhunk://#diff-a221f50cd7508ac36af65ce9b0956a32ef67caa47b9d37311989f9f72807a6bfR1-R77): Added tests for the `TrimExtraFields` function to ensure it correctly filters out unwanted fields from the YAML configuration.

### DIFF
--- a/clientutil/trimconfig.go
+++ b/clientutil/trimconfig.go
@@ -1,0 +1,28 @@
+// Copyright (c) The Thanos Authors.
+// Licensed under the Apache License 2.0.
+
+package clientutil
+
+import (
+	"github.com/pkg/errors"
+	"gopkg.in/yaml.v2"
+)
+
+func TrimExtraFields(confContentYaml []byte) ([]byte, error) {
+	var raw map[string]interface{}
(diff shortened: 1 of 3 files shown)


clientutil/trimconfig.go
```diff
@@ -0,0 +1,28 @@
...
+		"config": true,
+		"prefix": true,
+	}
+	filtered := make(map[string]interface{})
+	for key, value := range raw {
+		if allowed[key] {
+			filtered[key] = value
+		}
+	}
+	return yaml.Marshal(filtered)
```
MichaHoffmann, line 27: Isnt it kinda weird that we marshal it back to yaml just to parse it again?

Issue #169, opened 2025-03-07
Setup dependabot


bwplotka:
It would be epic to have periodic updates in this repo.

Help wanted!

bwplotka:
Although we have it https://github.com/thanos-io/objstore/blob/main/.github/dependabot.yml 🤔

bwplotka:
Ah it's broken 

![Image](https://github.com/user-attachments/assets/2b2e0a40-f240-4613-9cc8-cc8ca791e3b4)

PR #170, opened 2025-03-07
Fix dependabot.yml by removing invalid field.

bwplotka:
Fixes https://github.com/thanos-io/objstore/issues/169

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,7 +2,6 @@ version: 2
 updates:
   - package-ecosystem: "gomod"
     directory: "/"
-    vendor: false
     schedule:
-      interval: "weekly"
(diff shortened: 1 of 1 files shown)


saswatamcode (approved):

.github/dependabot.yml
```diff
@@ -2,7 +2,6 @@ version: 2
 updates:
```
saswatamcode, line 2: Looks like actions/cache version is outdated so CI fails, maybe we can add
```
- package-ecosystem: "github-actions"
    directory: "/"
    schedule: 
      interval: weekly
```

saswatamcode: Was added later, but need to merge this to unblock

PR #178, opened 2025-04-21
Feature: conditional upload API

### LINKED ISSUE #129: Feature request: conditions API
Firstly thank you for this library I've just started using it and the experience is really good.

Many object stores now support conditional gets and writes. The former allows better caching, and the latter allows developers to write data to an object if and only if the object hasn't been changed since a previous point in time, which forms the backbone for 'atomic' updates to datasets on object storage (a reference use case would be databricks delta format).

Generally object store providers either provide this functionality using a conditional header or payload containing an _etag_, a _version_ or _generation_ or even a last modified. 

Examples:
GCS: https://pkg.go.dev/cloud.google.com/go/storage#hdr-Conditions
Azure: https://learn.microsoft.com/en-us/rest/api/storageservices/specifying-conditional-headers-for-blob-service-operations

Note - (infamously?) **AWS** S3 doesn't support this. The standard response to this seems to be to [ask library users](https://delta-io.github.io/delta-rs/usage/writing/writing-to-s3-with-locking-provider/) to implement their own locking provider, and not address this in the library.

Here is an example of another library that has nice API for this: https://docs.rs/object_store/latest/object_store/struct.UpdateVersion.html

My thought would be that the ObjectAttributes API could include this metadata, and then I'm not sure how the API would be be changed conservatively to support this for Get or Upload; I quite like the minimal addition of an If(condition) receiver function that returns a wrapped object (similar to the cloud-storage library).

If this feature gets the maintainers' blessing I am happy to contribute pull requests for GCS and a suitable-for-testing purposes version for LocalFilesystem.

Thanks again.

alsenz:
Resolves #129 

### Adds support for conditional object uploads.

Extends `ObjectAttributes` with optional `ObjectVersion` (if supported), and `Upload(...)` with three new `ObjectUploadOption`, depending on provider support.
1. `IfMatch`: the object is written only if the existing object matches the provided version
2. `IfNotMatch`: the object is written only if the existing object does not match the provided version
3. `IfNotExists`: the object is written only if no object already exists for the provided key

Supports two forms of `ObjectVersion`:
1. Generational versions (i.e. increasing integers)
2. Etags

Not all providers support conditional write. The following providers are supported:
1. Filesystem: uses extended filesystem attributes. Only supported if the host system supports extended attributes.
2. InMem
3. GCS
4. Azure
5. S3: `IfNotMatch` is not yet supported (as it is not supported by AWS).

Clients can check conditional API by calling `SupportedObjectUploadOptions` on the `Bucket` interface.

### API changes

1. Added 3 new `ObjectUploadOption`s: `IfMatch`, `IfNotMatch` and `IfNotExists`
  1.1 Made `ObjectUploadOption` typed, repeating the existing `IterOption` pattern.
  1.2 Added three new parameter modifiers- `WithIfMatch`, `WithIfNotExists` and `WithIfNotMatch`, and `WithContentType` to support the previous un-typed `ObjectUploadOption`
5. Added optional `ObjectVersion` field to `ObjectAttributes`
7. Added acceptance tests for `IfMatch`, `IfNotMatch`, `IfNotExists` and for `Attributes` API
8. Added `Bucket` provider implementations for:
  1.1 filesystem: this now uses [Extended Attributes](https://en.wikipedia.org/wiki/Extended_file_attributes), if supported by the host system
  1.2 inmem
  1.3 GCS
  1.4 Azure
  1.5 S3, excepting `IfNoneMatch` which is not fully supported (by AWS) yet
  1.6 Wrappers
  
### Backwards compatibility

API changes are additive and should be backwards compatible.
- The `Upload` API changes use a trailing variadic parameter so should be backwards compatible
- The `ObjectAttributes` add new field so should be backwards compatible

- The `SupportedObjectUploadOptions` interface method potentially breaks existing `Bucket` implementations. Library users who implement their own providers will need to implement this method. Full or stub implementations are provided for all provider implementations in the libary.

### Dependencies

- Adds dependency on github.com/pkg/xattr, which is BSD-2-clause licensed. 

## Verification

Added acceptance tests for 
- `IfMatch`
- `IfNotMatch`
- `IfExists`
- Versions on Attributes

Did not run integration tests on other object store providers, as I do not have access to test environments, but these should be unchanged.


alsenz:
@bwplotka  - would you be the appropriate maintainer to submit this for a first review?

Greatly appreciated.

providers/filesystem/filesystem.go
```diff
@@ -246,29 +276,136 @@ func (b *Bucket) Exists(ctx context.Context, name string) (bool, error) {
...
+		swf, err = os.OpenFile(name, os.O_RDWR|os.O_CREATE|os.O_EXCL, 0666)
+		if err == nil {
+			return
+		} else if !errors.Is(err, fs.ErrExist) {
+			return
+		}
+	}
+}
+
+func tryOpenFile(name string, ifNotExists bool) (exists bool, err error) {
```
yuchen-db, line 291: does it leak file descriptor?

objstore.go
```diff
@@ -274,12 +260,130 @@ func applyUploadOptions(options ...UploadOption) uploadParams {
...
+		Type: IfNotMatch,
+		Apply: func(params *UploadObjectParams) {
+			params.Condition = ver
+			params.IfNotMatch = true
+		},
+	}
+}
+
+// ValidateUploadOptions ensures that only supported options are passed as options.
+func ValidateUploadOptions(supportedOptions []ObjectUploadOptionType, opts ...ObjectUploadOption) error {
```
yuchen-db, line 336: maybe we should also validate `IfNotExists` and `IfMatch/IfNotMatch` aren't used together

inmem.go
```diff
@@ -213,9 +220,36 @@ func (b *InMemBucket) Attributes(_ context.Context, name string) (ObjectAttribut
 }
 
 // Upload writes the file specified in src to into the memory.
-func (b *InMemBucket) Upload(_ context.Context, name string, r io.Reader, _ ...ObjectUploadOption) error {
+func (b *InMemBucket) Upload(_ context.Context, name string, r io.Reader, opts ...ObjectUploadOption) error {
 	b.mtx.Lock()
 	defer b.mtx.Unlock()
+
+	params := ApplyObjectUploadOptions(opts...)
```
yuchen-db, line 270: shall we validate the options here?

yuchen-db (approved):

anarcher:
What's the progress on this PR? Any chance it gets merged?

providers/filesystem/filesystem.go
alsenz: Agreed - fixed.

objstore.go
alsenz: Sounds sensible - done.

inmem.go
alsenz: Whoops - done!

alsenz:
Thanks @yuchen-db for your review. I agree with all your points and have pushed some changes accordingly. This is an old PR but hopefully we can get it across the line.

alsenz:
@anarcher this is an older PR that didn't get much attention, hopefully the maintainers will be able to find some time to review and accept. I have brought the feature branch back in line with main and responded to comments, so it should be ready to go.

RichardZhangRZ:
Would like to see this PR merged soon :)

GiedriusS:
I'll try to take a look no later than 2 months~ from now

providers/azure/azure.go
```diff
@@ -304,6 +307,13 @@ func (b *Bucket) IsAccessDeniedErr(err error) bool {
 	return bloberror.HasCode(err, bloberror.AuthorizationPermissionMismatch) || bloberror.HasCode(err, bloberror.InsufficientAccountPermissions)
 }
 
+// IsConditionNotMetErr returns true if the error was a blob error nad the code is ConditionNotMet, or BlobAlreadyExists.
```
GiedriusS, line 310: In the comment it is better not to write what is already in the code. Maybe this would be better:

```suggestion
// IsConditionNotMetErr returns true if the given conditions (e.g. the given ETag matches) were not met.
```

providers/filesystem/filesystem.go
```diff
@@ -246,29 +277,137 @@ func (b *Bucket) Exists(ctx context.Context, name string) (bool, error) {
...
+
+	if err := objstore.ValidateUploadOptions(b.SupportedObjectUploadOptions(), opts...); err != nil {
+		return err
+	}
+
 	if ctx.Err() != nil {
 		return ctx.Err()
 	}
 
 	file := filepath.Join(b.rootDir, name)
```
GiedriusS, line 319: Why can't we create an extended attribute on the same file?

providers/gcs/gcs.go
```diff
@@ -367,6 +399,18 @@ func (b *Bucket) IsAccessDeniedErr(err error) bool {
 	return false
 }
 
+// IsConditionNotMetErr returns true if the response status code was Precondition Failed or Not Modified.
```
GiedriusS, line 402: The same here: I suggest not adding into the comments what is already in the code.

providers/s3/s3.go
```diff
@@ -558,32 +561,47 @@ func (b *Bucket) Upload(ctx context.Context, name string, r io.Reader, opts ...o
...
+		// ensure we pin this number to four.
+		// TODO(bwplotka): Consider adjusting this number to GOMAXPROCS or to expose this in config if it becomes bottleneck.
+		NumThreads:  4,
+		ContentType: uploadOpts.ContentType,
+	}
+
+	if uploadOpts.IfNotExists {
+		putOpts.SetMatchETagExcept("*")
+	} else if uploadOpts.Condition != nil {
+		// If-None-Match with header values other than "*" is not supported by AWS yet.
```
GiedriusS, line 581: Should we error out on this condition? How the user would know about this? The user could pass something and it would do something different than what is expected so I would expect an error here.

GiedriusS:
@alsenz ping

alsenz:
@GiedriusS just seen this thanks for the review (been a while sorry).

I will resolve these tonight after work

providers/azure/azure.go
alsenz: Resolved.

providers/filesystem/filesystem.go
alsenz: This was written a while ago, so apologies if I'm a little hazy, but I'm fairly use we're using the swap-and-move idiom to prevent race conditions when the file is created but the xattrs aren't written, which would could cause versioning assumptions to go wrong- this relies on atomic move of course but that's already ensured by `SupportedObjectUploadOptions` - we don't do this code path if we don't have xattrs and atomic moves. 

To double check, I ran this through claude (which I try to avoid relying on but is a good double-check), it said:
 ▎ The xattr is set on the swap file before the atomic os.Rename. Since os.Rename on Unix is atomic and xattrs are stored as inode metadata, they travel with the file on rename. Setting the xattr on file after the    
  rename would introduce a race: another goroutine could call checkConditions in that window and see the file without its version data.

providers/gcs/gcs.go
alsenz: Agreed - will fix in next commit

providers/s3/s3.go
alsenz: Absolutely - yes, good catch! Will fix in next commit.

alsenz:
@GiedriusS  -- PR comments addressed - haven't resolved the comments yet until you take a look. my best.

README.md
```diff
@@ -59,7 +59,10 @@ type Bucket interface {
 
 	// Upload the contents of the reader as an object into the bucket.
 	// Upload should be idempotent.
-	Upload(ctx context.Context, name string, r io.Reader, opts ...ObjectUploadOption) error
+	Upload(ctx context.Context, name string, r io.Reader, options ...ObjectUploadOption) error
+
```
GiedriusS: Some random whitespace issues here.

objstore.go
```diff
@@ -274,12 +260,139 @@ func applyUploadOptions(options ...UploadOption) uploadParams {
...
+const (
+	ContentType ObjectUploadOptionType = iota
+	IfNotExists
+	IfMatch
+	IfNotMatch
+)
+
+// ObjectUploadOption configures UploadObjectParams.
+type ObjectUploadOption struct {
+	Type  ObjectUploadOptionType
```
GiedriusS: I don't like that these two are disjointed now and users can misuse this ie create some `ObjectUploadOption` that doesn't do what `Type` says it should do. Should we make `Type` and `Apply` hidden i.e. lowercase so that users wouldn't be able to do that?

providers/filesystem/filesystem.go
```diff
@@ -246,29 +277,137 @@ func (b *Bucket) Exists(ctx context.Context, name string) (bool, error) {
...
+		if err == nil {
+			return
+		} else if !errors.Is(err, fs.ErrExist) {
+			return
+		}
+	}
+}
+
+func tryOpenFile(name string, ifNotExists bool) (exists bool, err error) {
+	// First try to open the file with exclusive create, then truncate if permitted
```
GiedriusS: Where truncation happens here? 🤔

providers/filesystem/filesystem.go
```diff
@@ -246,29 +277,137 @@ func (b *Bucket) Exists(ctx context.Context, name string) (bool, error) {
...
+		return err
+	}
+
+	if xattr.XATTR_SUPPORTED {
+		h := sha256.New()
+		writer := io.MultiWriter(swf, h)
+		if _, err := io.Copy(writer, r); err != nil {
+			return errors.Wrapf(err, "copy to %s", swap)
+		}
+		// Write the checksum into an xattr
```
GiedriusS: I suggest removing self-explanatory comments like this.

providers/filesystem/filesystem.go
```diff
@@ -246,29 +277,137 @@ func (b *Bucket) Exists(ctx context.Context, name string) (bool, error) {
...
 	}
-	defer errcapture.Do(&err, f.Close, "close")
 
-	if _, err := io.Copy(f, r); err != nil {
-		return errors.Wrapf(err, "copy to %s", file)
+	if err := b.checkConditions(name, params, exists); err != nil {
+		return err
+	}
+
+	if xattr.XATTR_SUPPORTED {
```
GiedriusS, line 329: If this is false then where are we writing to `swf`?

objstore.go
```diff
@@ -274,12 +260,139 @@ func applyUploadOptions(options ...UploadOption) uploadParams {
...
+		Apply: func(params *UploadObjectParams) {
+			params.Condition = ver
+			params.IfNotMatch = true
+		},
+	}
+}
+
+// ValidateUploadOptions ensures that only supported options are passed as options, and that options used simultaneously are valid.
+func ValidateUploadOptions(supportedOptions []ObjectUploadOptionType, opts ...ObjectUploadOption) error {
+
```
GiedriusS: Typically we don't put a empty line just after the definition.

providers/filesystem/filesystem.go
```diff
@@ -246,29 +277,137 @@ func (b *Bucket) Exists(ctx context.Context, name string) (bool, error) {
...
+		return
+	}
+	err = f.Close()
+	return
+}
+
 // Upload writes the file specified in src to into the memory.
-func (b *Bucket) Upload(ctx context.Context, name string, r io.Reader, _ ...objstore.ObjectUploadOption) (err error) {
+func (b *Bucket) Upload(ctx context.Context, name string, r io.Reader, opts ...objstore.ObjectUploadOption) (err error) {
+
```
GiedriusS: Same here.

GiedriusS (commented):
Thank you, a few more suggestions

README.md
alsenz: Will resolve.

objstore.go
alsenz: No fair enough, although I think there may be other places where symbols are overexposed maybe. Either way, easy change to make and very happy to do.

providers/filesystem/filesystem.go
alsenz: No idea. I think it's a comment typo rather than code typo, re-readin gthe whole logic.

alsenz: I'll just remove the comment fo rnow

providers/filesystem/filesystem.go
alsenz: Will remove

providers/filesystem/filesystem.go
alsenz: This is a bug- thanks -  think the fix is a few lines, will fix.

objstore.go
alsenz: Will do

providers/filesystem/filesystem.go
alsenz: :thumbs-up:

alsenz:
Thanks for the second pass @GiedriusS  - will apply changes tonight and push.

tracing/opentracing/opentracing.go
```diff
@@ -142,6 +146,10 @@ func (t TracingBucket) IsAccessDeniedErr(err error) bool {
 	return t.bkt.IsAccessDeniedErr(err)
 }
 
+func (t TracingBucket) IsConditionNotMetErr(err error) bool {
+	return t.bkt.IsAccessDeniedErr(err)
```
GiedriusS: Should this be IsConditionNotMetErr?

providers/filesystem/filesystem.go
```diff
@@ -246,29 +277,137 @@ func (b *Bucket) Exists(ctx context.Context, name string) (bool, error) {
...
+	return
+}
+
+// checksum reads an X-Attr for the checksum property if it was written with the file, or empty string if it was not.
+// Does not check if X-Attrs are supported on the host - this must be done by the caller.
+func (b *Bucket) checksum(name string) (string, error) {
+	file := filepath.Join(b.rootDir, name)
+	bytes, err := xattr.Get(file, xAttrKey)
+	if err != nil {
+		return "", err // Legacy filesystem buckets would just return empty string for the version (until objects updated).
```
GiedriusS: According to the comment, shouldn't `nil` error be returned?

providers/filesystem/filesystem.go
```diff
@@ -246,29 +277,137 @@ func (b *Bucket) Exists(ctx context.Context, name string) (bool, error) {
 	return !info.IsDir(), nil
 }
 
-// Upload writes the file specified in src to into the memory.
-func (b *Bucket) Upload(ctx context.Context, name string, r io.Reader, _ ...objstore.ObjectUploadOption) (err error) {
+func openSwap(name string) (swf *os.File, err error) {
+	for {
```
GiedriusS: Why is this a loop? If creating this fails for some reason other than `fs.ErrExist` then this will loop indefinitely 🤔

providers/filesystem/filesystem.go
```diff
@@ -246,29 +277,137 @@ func (b *Bucket) Exists(ctx context.Context, name string) (bool, error) {
...
+	f, err = os.OpenFile(name, flags, 0666)
+	if errors.Is(err, fs.ErrExist) && !ifNotExists {
+		exists = true
+		flags = os.O_RDWR | os.O_CREATE | os.O_APPEND
+		f, err = os.OpenFile(name, flags, 0666)
+	}
+	if f == nil {
+		return
+	}
+	err = f.Close()
```
GiedriusS: This is immediately closed so maybe we can just call `os.Stat` instead?

GiedriusS (approved):
Fixed a few nits from my review. Hope that's OK. Merging this. Thank you for your contribution! I will try to use it in the Thanos parquet converter and there might be a few more things that will need to be changed but overall I think it's good!

PR #181, opened 2025-05-27
GCS: Fix issue where `Exists` returns an error when the object does not exist with GCS client v1.51+

charleskorn:

## Changes

v1.51 of the GCS client includes https://github.com/googleapis/google-cloud-go/pull/11519, which changes the behaviour of `Attrs()` to return a wrapped `ErrObjectNotExist` (rather than `ErrObjectNotExist` itself) when the object does not exist. This breaks the error check done in `Bucket.Exists()` to detect when the object does not exist.

### DIFF
--- a/providers/gcs/gcs.go
+++ b/providers/gcs/gcs.go
@@ -325,7 +325,7 @@ func (b *Bucket) Handle() *storage.BucketHandle {
 func (b *Bucket) Exists(ctx context.Context, name string) (bool, error) {
 	if _, err := b.bkt.Object(name).Attrs(ctx); err == nil {
 		return true, nil
-	} else if err != storage.ErrObjectNotExist {
+	} else if b.IsObjNotFoundErr(err) {
 		return false, err
 	}
 	return false, nil


providers/gcs/gcs.go
```diff
@@ -325,7 +325,7 @@ func (b *Bucket) Handle() *storage.BucketHandle {
 func (b *Bucket) Exists(ctx context.Context, name string) (bool, error) {
 	if _, err := b.bkt.Object(name).Attrs(ctx); err == nil {
 		return true, nil
-	} else if err != storage.ErrObjectNotExist {
+	} else if b.IsObjNotFoundErr(err) {
```
aknuds1, line 328: Shouldn't this be the inverse?

```suggestion
	} else if !b.IsObjNotFoundErr(err) {
```

aknuds1 (commented):
Please see question.

providers/gcs/gcs.go
charleskorn: 🤦 Good catch, fixed.

aknuds1 (approved):
Could this PR also include a fix, to prevent a future regression?

charleskorn:
> Could this PR also include a test, to prevent a potential future regression?

This would be covered by [the existing acceptance tests](https://github.com/thanos-io/objstore/blob/a0136a6f898d43ec0e654a98ac322554f2ae259b/testing.go#L83), once this project upgrades to one of the affected GCS client versions.

aknuds1:
@charleskorn OK, got it, thanks!

charleskorn:
Any chance I could get a review on this from a maintainer? Would be great to get this merged.

JoaoBraveCoding (commented):
Looks good can you rebase then I'm fine with giving it the approve

PR #182, opened 2025-06-03
download: dont cleanup directory after download failed

MichaHoffmann:
DownloadDir has the option to ignore paths that we have already on disk. In theory that makes downloads resumable, but in practice this doesnt work because we delete everything if one download fails. This means if one chunk in 1TiB block fails to download for whatever reason we start all over.

### DIFF
--- a/objstore.go
+++ b/objstore.go
@@ -14,7 +14,6 @@ import (
 	"path/filepath"
 	"slices"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/efficientgo/core/logerrcapture"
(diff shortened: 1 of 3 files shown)


objstore.go
```diff
@@ -442,7 +441,11 @@ func DownloadFile(ctx context.Context, logger log.Logger, bkt BucketReader, src,
 }
 
 // DownloadDir downloads all object found in the directory into the local directory.
-func DownloadDir(ctx context.Context, logger log.Logger, bkt BucketReader, originalSrc, src, dst string, options ...DownloadOption) error {
+func DownloadDir(ctx context.Context, logger log.Logger, bkt BucketReader, src, dst string, options ...DownloadOption) error {
+	return downloadDir(ctx, logger, bkt, src, src, dst, options...)
+}
+
+func downloadDir(ctx context.Context, logger log.Logger, bkt BucketReader, originalSrc, src, dst string, options ...DownloadOption) error {
```
aknuds1, line 448: I'm wondering first of all whether `downloadDir` shouldn't return immediately if `ctx.Err()` is non-nil. The reason is that `downloadDir` can be called recursively with a context from `errgroup.WithContext`, that gets canceled when one or more of the goroutines fail.

objstore.go
```diff
@@ -474,34 +471,14 @@ func DownloadDir(ctx context.Context, logger log.Logger, bkt BucketReader, origi
...
 			}
-			if err := DownloadFile(ctx, logger, bkt, name, dst); err != nil {
-				return err
-			}
-
-			m.Lock()
-			downloadedFiles = append(downloadedFiles, dst)
-			m.Unlock()
-			return nil
+			return DownloadFile(ctx, logger, bkt, name, dst)
```
aknuds1, line 474: Does `DownloadFile` respect cancellation via `ctx`?

aknuds1 (commented):
Leaving some comments. I lack context to judge whether the change is going to break anything though. Let me ask the Mimir team.

PR #183, opened 2025-06-27
[deps] Update minio-go and github actions

coleenquadros:
Need to update objstore minio-go before Thanos -> https://github.com/thanos-io/thanos/issues/8201 


saswatamcode (approved):
generally LGTM, but you might need to bump golangci-lint with bingo!

.golangci.yml
```diff
@@ -41,17 +36,16 @@ linters:
...
     - unused
-    - varcheck
     - exportloopref
     - promlinter
 
 linters-settings:
   errcheck:
-    exclude: ./.errcheck_excludes.txt
+    exclude-functions:
+      - ./.errcheck_excludes.txt
```
saswatamcode, line 48: Looks like this can no longer be a txt file, but actually the contents of the file in YAML list form,
      exclude-functions:
        - (github.com/go-kit/log.Logger).Log
        - fmt.Fprintln
        - fmt.Fprint

saswatamcode: As per https://golangci-lint.run/usage/linters/#errcheck

.github/workflows/docs.yaml
```diff
@@ -18,11 +18,11 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Install Go
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@0a12ed9d6a96ab950c8f026ed9f722fe0da7ef32 # v5.0.2
```
saswatamcode, line 21: Why are we pinning these to commit, we could just use latest release right?

.go-version
```diff
@@ -1 +1 @@
-1.21.x
+1.24
```
saswatamcode, line 1: Maybe?
```suggestion
1.24.x
```

saswatamcode (approved):
LGTM, just a final couple comments!

saswatamcode (approved):
LGTM

PR #218, opened 2025-09-03
chore: add size to iter with attributes

MichaHoffmann:

### DIFF
--- a/objstore.go
+++ b/objstore.go
@@ -284,6 +284,7 @@ type ObjectAttributes struct {
 
 type IterObjectAttributes struct {
 	Name         string
+	Size         int64
 	lastModified time.Time
 }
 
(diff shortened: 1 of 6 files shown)


GiedriusS (commented):
Should we name it SizeBytes to avoid any kind of confusion over what are the units?

MichaHoffmann:
> Should we name it SizeBytes to avoid any kind of confusion over what are the units?

Yeah we can do that! Ill also implement it for the other providers a bit later!

providers/azure/azure.go
```diff
@@ -232,6 +232,7 @@ func (b *Bucket) IterWithAttributes(ctx context.Context, dir string, f func(attr
 			for _, blob := range resp.Segment.BlobItems {
 				attrs := objstore.IterObjectAttributes{
 					Name: *blob.Name,
+					Size: *blob.Properties.ContentLength,
```
saswatamcode, line 235: SizeBytes would be preferable, I think!

objstore.go
```diff
@@ -284,6 +284,7 @@ type ObjectAttributes struct {
 
 type IterObjectAttributes struct {
 	Name         string
+	Size         int64
```
GiedriusS, line 287: Can the size be negative? Maybe let's use uint64.

MichaHoffmann: Maybe as a canary for "this provider does not expose a size in iter"; idk if there is such a provider though

PR #247, opened 2026-01-07
feat(tlsConfig): Add possiblity to trust provided CA and host at the same time

davinkevin:

## Changes

This change introduces a new flag in the TLSConfig to leverage both `CACert` provided by end-users and System certificates from the host. 

Until now, it was not possible to do both at the same time, leading to complex configuration when cert chains are too long (k8s size limits for configMaps) or when we share certificates with other application using an "additive system" (as the opposite of the code before this PR). 

## Verification

I tested the code locally, but I'm open to more testing as my client was built into `mimir`, so maybe not good enough. 

**NOTE** I've opened a PR as this change is so small, it would be simpler to discuss around a PR than an issue. If it's ok for you, then I'll modify the changelog and other mandatory items.

Thank you in advance.

### DIFF
--- a/exthttp/tlsconfig.go
+++ b/exthttp/tlsconfig.go
@@ -14,6 +14,8 @@ import (
 type TLSConfig struct {
 	// The CA cert to use for the targets.
 	CAFile string `yaml:"ca_file"`
+	// Trust RootCAs provided by the host
+	TrustRootCA bool `yaml:"trust_root_ca"`
 	// The client cert file for the targets.
 	CertFile string `yaml:"cert_file"`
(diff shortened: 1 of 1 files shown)


pio-kol (approved):

davinkevin:
ping @MichaHoffmann @matej-g @bwplotka; I don't like to ping like this, but I had no answer for several month and I'd like to progress on that one. 

If you agree with the modification, I could progress on many aspects (changelog, tests, and others if needed). 

Let me know! 

Thank you 🙏

pio-kol (approved):

PR #257, opened 2026-02-19
gcs: add endpoint configuration option

### LINKED ISSUE #7862: Support setting endpoint for Google Cloud Storage (GCS) in Thanos
### Is your proposal related to a problem?

We need to override the Google API endpoint used by Thanos for GCS storage to ensure usage of [GCP regional endpoints](https://cloud.google.com/storage/docs/regional-endpoints).  

### Describe the solution you'd like

Proposal for a solution:

1. Add an option which can be used to configure the endpoint for the GCP storage client
2. Enhance the GCS client to set it using the option.WithEndpoint(...): https://github.com/grafana/thanos/blob/main/scripts/cfggen/main.go
3. It seems thanos-io/objstore is also used to support common actions with GCS. The functionality should also be added here: https://github.com/thanos-io/objstore/blob/main/providers/gcs/gcs.go

### Describe alternatives you've considered
/

### Additional context
/

rohansood10:
## Problem

Thanos currently does not support configuring a custom API endpoint for Google Cloud Storage. This prevents using [GCS regional endpoints](https://cloud.google.com/storage/docs/regional-endpoints) or GCS-compatible storage backends.

Fixes: thanos-io/thanos#7862

## Solution

Added an `endpoint` field to the GCS `Config` struct. When set, it is passed as `option.WithEndpoint()` when creating the GCS storage client.

### Configuration example

```yaml
type: GCS
config:
  bucket: my-bucket
  endpoint: https://storage.europe-west3.rep.googleapis.com
```

## Changes

- **providers/gcs/gcs.go**: Added `Endpoint` field to `Config` struct and wired it through `NewBucketWithConfig` via `option.WithEndpoint()`
- **providers/gcs/gcs_test.go**: Added tests for endpoint config parsing and bucket creation with custom endpoint

## Notes

- Follows the same pattern used by the S3 provider for custom endpoints
- When `endpoint` is empty (default), behavior is unchanged
- Compatible with both HTTP and gRPC client modes

### DIFF
--- a/README.md
+++ b/README.md
@@ -392,6 +392,7 @@ config:
     disable_compression: false
   chunk_size_bytes: 0
   max_retries: 0
+  endpoint: ""
 prefix: ""
 ```
 
--- a/providers/gcs/gcs.go
+++ b/providers/gcs/gcs.go
@@ -59,6 +59,11 @@ type Config struct {
 	// Overrides the default gcs storage client behavior if this value is greater than 0.
 	// Set this to 1 to disable retries.
 	MaxRetries int `yaml:"max_retries"`
+
+	// Endpoint specifies the GCS API endpoint to use.
+	// This can be used to point to GCS regional endpoints or GCS-compatible storage backends.
+	// See https://cloud.google.com/storage/docs/regional-endpoints for more details.
+	Endpoint string `yaml:"endpoint"`
 }
 
 // Bucket implements the store.Bucket and shipper.Bucket interfaces against GCS.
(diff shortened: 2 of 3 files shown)


rohansood10:
@bwplotka @JoaoBraveCoding @GiedriusS hey! Fixed the DCO sign-off issue and rebased onto main. Would appreciate a review when you get a chance - this adds a configurable endpoint option for GCS, useful for emulator testing and GCS-compatible storage.

PR #265, opened 2026-07-29
s3: tidy up transport construction

bboreham:

## Changes

Only build the default transport if none is set in config.
This is what the comment already said; it confused me that the code was doing extra work.

## Verification

It's simple enough to eyeball.  Hopefully there are tests.

### DIFF
--- a/providers/s3/s3.go
+++ b/providers/s3/s3.go
@@ -247,12 +247,14 @@ func NewBucketWithConfig(logger log.Logger, config Config, component string, wra
 	// Check if a roundtripper has been set in the config
 	// otherwise build the default transport.
 	var tpt http.RoundTripper
-	tpt, err := exthttp.DefaultTransport(config.HTTPConfig)
-	if err != nil {
-		return nil, err
-	}
 	if config.HTTPConfig.Transport != nil {
 		tpt = config.HTTPConfig.Transport
+	} else {
+		var err error
+		tpt, err = exthttp.DefaultTransport(config.HTTPConfig)
+		if err != nil {
+			return nil, err
+		}
 	}
 	if wrapRoundtripper != nil {
 		tpt = wrapRoundtripper(tpt)


JoaoBraveCoding (commented):
Personally I tend to prefer the default & overwrite approach vs using `else` as it makes the code easier to read/maintain. From my POV I would rather see the comment change.

bboreham:
Just to check: you get that it's doing work - allocating objects, reading CA & cert files - which is all wasted in the case that is discovered on the next line.

JoaoBraveCoding (approved):
Thank you for the follow-up, yes I do understand that. I mainly said my 2 cents, from a POV of readability and having in mind that this only happens at creation, not on the hot path. 

Regardless, here is a benchstat comparing the two to sustain your change if anyone checks the PR in the future.

```
                                      │     old.txt     │               new.txt               │
                                      │     sec/op      │   sec/op     vs base                │
TransportConstruction/no_tls_files-22     335.300n ± 1%   1.279n ± 2%  -99.62% (p=0.000 n=15)
TransportConstruction/with_ca_file-22   10237.000n ± 2%   1.280n ± 2%  -99.99% (p=0.000 n=15)
geomean                                     1.853µ        1.279n       -99.93%

                                      │   old.txt    │                  new.txt                  │
                                      │     B/op     │     B/op      vs base                     │
TransportConstruction/no_tls_files-22   1.188Ki ± 0%   0.000Ki ± 0%  -100.00% (p=0.000 n=15)
TransportConstruction/with_ca_file-22   6.594Ki ± 0%   0.000Ki ± 0%  -100.00% (p=0.000 n=15)
geomean                                 2.798Ki                      ?                       ¹ ²
¹ summaries must be >0 to compute geomean
² ratios must be >0 to compute geomean

                                      │  old.txt   │                 new.txt                 │
                                      │ allocs/op  │ allocs/op   vs base                     │
TransportConstruction/no_tls_files-22   5.000 ± 0%   0.000 ± 0%  -100.00% (p=0.000 n=15)
TransportConstruction/with_ca_file-22   64.00 ± 0%    0.00 ± 0%  -100.00% (p=0.000 n=15)
geomean                                 17.89                    ?                       ¹ ²
¹ summaries must be >0 to compute geomean
² ratios must be >0 to compute geomean
```

From a curiosity POV, would be interested in the real world impact of this change in your systems. Thank you for the contribution!

bboreham:
> From a curiosity POV, would be interested in the real world impact of this change in your systems. 

None; it was just wrong.  It matters a little more if you add #271.

PR #271, opened 2026-08-24
feat(http): Add connection metrics

bboreham:

## Changes

Add metrics `net_conntrack_dialer_conn_*_total` for every connection attempted, failed, closed, etc.

Conntrack metrics have a `dialer_name` label that can be useful to see connections for different reasons. 
This PR sets it to `"objstore"`, and there is a helper `conntrack.DialNameToContext` which lets you pass the name via context from the point where the dial is triggered.

## Commentary

I considered adding something to put in `dialer_name` as a parameter to `DefaultTransport`, but this seems like a rather invasive change.

## Verification

I have been using this with Loki, which uses `thanos-io/objstore` for blob access.

### DIFF
--- a/exthttp/transport.go
+++ b/exthttp/transport.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"time"
 
+	conntrack "github.com/mwitkow/go-conntrack"
 	"github.com/prometheus/common/model"
 )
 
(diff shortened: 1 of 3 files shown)


exthttp/transport.go
```diff
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"time"
 
+	conntrack "github.com/mwitkow/go-conntrack"
```
JoaoBraveCoding, line 11: Could yo run `make deps` to make conntrack a direct dependency.

exthttp/transport.go
```diff
@@ -52,14 +53,19 @@ func DefaultTransport(config HTTPConfig) (*http.Transport, error) {
...
-			DualStack: true,
-		}).DialContext,
+	dialFn := (&net.Dialer{
+		Timeout:   30 * time.Second,
+		KeepAlive: 30 * time.Second,
+		DualStack: true,
+	}).DialContext
+	// conntrack exports metrics for every connection attempted, failed, etc.
+	dialFn = conntrack.NewDialContextFunc(
+		conntrack.DialWithDialContextFunc(dialFn),
```
JoaoBraveCoding, line 63: Doesn't this add more value than just having `dialer_name="default"`? It also would be an entry point to the ability of overwriting with `DialNameToContext` WDYT?
```suggestion
		conntrack.DialWithDialContextFunc(dialFn),
		// Can be overwritten with conntrack.DialNameToContext.
		conntrack.DialWithName("objstore"),
```

